Issue #155, opened 2014-08-26
LifecycleManagerStarter action does not start LifecycleManager


romikk:
LifecycleManagerStarter does not call `start()` method on `LifecycleManager`
In fact it does nothing except retrieving it from injector:

``` java
public class LifecycleManagerStarter implements PostInjectorAction {
    public void call(Injector injector) {
        manager = injector.getInstance(LifecycleManager.class);
    }

    public void close() {
        if (manager != null) {
            manager.close();
        }
    }
}
```

Also `close()` method does not make any sense since it is not implementing any interface, and it's unclear how it will be called.

PR #156, opened 2014-08-26
#155 starting LifecycleManager

romikk:
LifecycleManagerStarter action does not start LifecycleManager #155


asafm:
@romikk What's the use case for this class LifecycleManagerStarer? I cant' seem to find a any documenation on it

PR #179, opened 2015-03-03
Support for Java 8

rdegnan:
Upgrade dependencies i.e. asm and library to support Java 8


ScottMansfield:
This update really didn't fix the problem. That's still an ASM jar from 2012 and doesn't support java 8. You need to update to org.ow2.asm:asm-all:5.0.3 and upgrade to org.apache.xbean:xbean-finder:4.1 to actually work.

elandau:
Since 1.7.x Governator was upgraded to Guice 4 and java8 compatible version of asm.

PR #186, opened 2015-03-25
Governator 1.4 cleanup

elandau:
* Deprecate AutoBindSingleton, AutoBind, AutoBindProvider, Configuration, ConfigurationVariable, PreConfiguration, ConfigurationProvider
* Introduce LifecycleAnnotationProcessor as pluggable abstraction so LifecycleManager can be more easily customized
* Deprecate general purpose bindings in BootstrapModule/BootstrapBinder.  These should be done on regular guide modules
* Make ClasspathScanner an abstract class so it can be more easily customized without depending on XBeanFinder.
* Main injector now created on LifecycleInjector.build().  All other injector creation methods have been deprecated.
* Break out archaius into a separate subproject so it is no longer a dependency of governator-core


governator-core/src/main/java/com/netflix/governator/annotations/AutoBind.java
```diff
@@ -27,6 +27,8 @@
  * Annotation binding that combines with Governator's classpath scanning and a bound
  * {@link AutoBindProvider} to automatically/programmatically bind fields and constructor/method
  * arguments
+ * 
+ * @deprecated  AutoBind is deprecated in favor of using Guice modules
  */
 @Documented
```
brharrington: It is probably a good idea to use the `@Deprecated` annotation as well to guarantee a warning will be shown and also improve detection of deprecation with other tools.

governator-core/src/main/java/com/netflix/governator/guice/InternalLifecycleModule.java
```diff
@@ -40,52 +36,42 @@
...
-                        new InjectionListener<T>() {
-                            @Override
-                            public void afterInjection(T obj) {
-                                processInjectedObject(obj, type);
-                            }
+    public static class LifecycleInjectionListener implements TypeListener {
+
+        // this really serves as a Set purpose.
+        // put dummy boolean as Map value.
+        // value is really not important here.
```
brharrington: Is this comment just out of date? I'm guessing a map was being used as set in some previous version.

governator-core/src/main/java/com/netflix/governator/lifecycle/DefaultClasspathScanner.java
```diff
@@ -0,0 +1,166 @@
...
+    public Set<Field> getFields()
+    {
+        return fields;
+    }
+
+    protected void doScanning(Collection<String> basePackages, Collection<Class<? extends Annotation>> annotations, Set<Class<?>> localClasses, Set<Constructor> localConstructors, Set<Method> localMethods, Set<Field> localFields)
+    {
+        if ( basePackages.size() == 0 )
+        {
+            log.warn("No base packages specified - no classpath scanning will be done");
```
brharrington: As I typically want to disable classpath scanning, I would prefer not getting a warning about it. That said, it looks like this change splits out the interface so I can easily pass in a no-op implementation.

PR #199, opened 2015-05-20
Simplified implementation of lifecycle annotations

elandau:
This PR provides a much simpler implementation of Governator functionality using standard guice modules and a lightweight Injector lifecycle manager.  The LifecycleInjector is created using the utility class Governator which mirrors {@Guice}'s createInjector methods.   

For example,

```java
InjectorLifecycle injector = Governator.createInjector(new LifecycleModule() /*, other modules to add to injector */);
injector.awaitTermination();
```

Support for Governator's bootstrap functionality is done as a utility class for creating modules.  Multiple bootstrap annotated classes may be added to the same injector in addition to other tranditional modules.  
```java
InjectorLifecycle injector = Governator.createInjector(
   new LifecycleModule(), 
   ModulesEx.fromClass(MyMainClass.class), 
   ModulesEx.fromClass(MyAdditionalMainClass.class));
```

A complete application can look like this,

```java
@CustomBootstrapConfiguration
@OtherBootstrapConfiguration
public static class MyMainClass extends AbstractModule {
    @Singleton
    public static class PostInjectorCreation extends DefaultLifecycleListener {
        @Override
        public void onReady() {
            System.out.println("Application bootstrapped");
        }
    }

    @Override
    protected void configure() { 
        install(new LifecycleModule());
        install(new ShutdownHookModule());
        install(new ServiceLoaderModuleBuilder().loadModules(MyPluginModule.class));

        Multibinder.newSetBinder(binder(), LifecycleListener.class).addBinding().to(PostInjectorCreation.class);
        // application bindings go here
    }

    public static void main(String args[]) {
        Governator.createInjector(ModulesEx.fromClass(MyMainClass.class)).awaitTermination();
    }
}
```

BootstrapModule and all other Governator classes are no longer needed.  Specifically, all configuration classes can be deprecated in favor of the Archaius2 guice module which provides configuration binding and configuration proxy creation from interfaces.  Classpath scanning will be provided in the future as a guice module.


governator-core/src/main/java/com/netflix/governator/LifecycleModule.java
```diff
@@ -0,0 +1,160 @@
...
+ *    }
+ *    
+ *    @PreDestroy
+ *    public void shutdown() {
+ *    }
+ * }
+ * }
+ * </pre>
+ * 
+ * To use simply add LifecycleModule to guice when creating the injector
```
qiangdavidliu: You'll need to update the comments now that this is installed by default.

governator-core/src/main/java/com/netflix/governator/LifecycleListener.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Contract for a callback to be invoked once the Injector is shut down.
+ * LifecycleListener should be reserved for framework use only.  All application 
+ * code should use @PreDestroy, which is triggered via a LifecycleListener
+ * enabled by installing {@link LifecycleModule}.
+ * 
+ * @author elandau
+ */
+public interface LifecycleListener {
```
qiangdavidliu: Is package private here possible?

PR #216, opened 2015-06-19
First pass at profile and condition based automatic module loading

elandau:
This PR provides automatic module loading using plugable module finders combined with conditional loading and optional overrides.  The power of AutoModuleBuilder comes into play when taking a complex system of modules and running it within different profiles where bindings must be overridden or extended based on various conditions such as properties, environment, operating system and existence of jars on the class path.
 
Using the feature code only needs to specify modules for its own bindings and defers loading of dependent modules based on the environment.  Implementations and overrides are loaded simply by adding jars to the classpath.  

Note that the output of the builder is a standard Guice module.  It is therefore possible to install multiple AutoModuleBuilder constructed modules as well as any other type of module.
 
Usage,
```java
      // This is your application main
      public static void main(String args[]) {
          Governator.createInjector( 
              AutoModuleBuilder
                  .forModule(new ApplicationModule())
                  .withProfile("production")
                  .build()
              .awaitTermination()
      }
      
      // Module to capture all the bindings for an application
      public static class ApplicationModule extends AbstractModule {
          protected void configure() {
              install(new SomeLibraryModule());
              ...
          }
      }
      
      // This module can be picked up by classpath scanning or via the service loader
      // Since the conditions match this module will be installed as an override.
      @ConditionalOnProfile({"production"})
      @OverrideModule(SomeLibraryModule.class)
      public static class SomeLibraryModuleOverride extends AbstractModule {
      }
   }
```

The above is equivalent to,
```java
      public static void main(String args[]) {
          Governator.createInjector( 
              Modules.override(new ApplicationModule())
                     .with(new SomeLibraryModuleOverride());
      }
   }
```
While the above example may not be too impressive this functionality is extremely helpful in an application with hundreds of modules, some transitive, where overrides and additional functionality is desired when running in different environments.

# Module discovery

Module discovery is handled via the ModuleListProvider abstraction.  By default modules are loaded via the JDK ServiceLoader by searching for metadata for com.google.inject.Module.   Alternative ModuleListProvider implementations can discovery modules via classpath scanning or from a fixed list of known modules classes.

# Supported conditions
* ConditonalOnBinding - A binding must exist
* ConditionalOnClass - A class must be on the classpath
* ConditionalOnEnvironment - An environment property must have value
* ConditionalOnJUnit - Running inside of JUnit
* ConditionalOnMacOS - Running on a MacOS
* ConditionalOnMissingBinding - A binding must not already exist
* ConditionalOnMissingClass - A class must not be on the classpath
* ConditionalOnMissingModule - A module must not have been installed
* ConditionalOnModule - A module must have been installed
* ConditionalOnProfile - Must have specified active profile
* ConditionalOnProperty - A pluggable configuration property must be set to a value.
* ConditionalOnSystem - A system property must be set to a value
* ConditionalOnTestNG - Running inside of TestNG

# Configuration based conditions

Several 'configuration' based conditions exist to control loading of modules from either System, Environment, or provided properties.  External configuration can be provided to the AutoModuleBuilder via the PropertySource interface.  By default PropertySource uses system and environment properties.

```java
Properties prop = new Properties();
prop.setProperty("foo", "bar)";

Governator.createInjector( 
    AutoModuleBuilder
        .forModule(new ApplicationModule())
        .withProfile("production")
        .withBootstrap(PropertiesConfig.toModule(prop))
        .build()
        .awaitTermination()

```


governator-core/src/main/java/com/netflix/governator/auto/ServiceLoaderModuleProvider.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public class ServiceLoaderModuleProvider implements ModuleListProvider {
+
+    @Override
+    public List<Module> get() {
+        List<Module> modules = new ArrayList<>();
+        Iterator<Module> iter = ServiceLoader.load(Module.class).iterator();
+        while (iter.hasNext()) {
+            Module module = iter.next();
+            if (null != module.getClass().getAnnotation(ConditionalOnProfile.class)) {
```
brharrington: Why is this specific to the profile condition? From the module discovery section in the PR desc I would have assumed it would work for all supported conditions.

elandau: I wanted to make sure that these modules at least to target a profile.  Alternatively I could either just make sure that there's at least one conditional.  It may also be desirable to restrict modules considered for auto loading based on package.

PR #217, opened 2015-06-26
Various fixes to AutoModule

elandau:


governator-core/src/main/java/com/netflix/governator/Governator.java
```diff
@@ -73,7 +73,9 @@ protected void configure() {
                 manager.notifyShutdown();
             }
             catch (Exception e2) {
-                
+            }
+            finally {
+                System.exit(-1);
```
brharrington: Just curious, what is the rationale for exiting the jvm? If used as to create an independent LifecycleInjector in a test for example that seems a bit extreme.

elandau: You are correct.  This should never have been pushed.

PR #229, opened 2015-08-06
Remove dependency on xbean-finder

jkschneider:
Perform annotation classpath scanning with a vanilla ASM ClassVisitor.

Further work could be done in the future that reads class bytecodes directly, with a potential performance improvement.  See similar attempts at https://github.com/lukehutch/fast-classpath-scanner and https://github.com/rmuller/infomas-asl/tree/master/annotation-detector.  This would remove even the ASM dependency at the cost of having to update with future changes in bytecodes.

Removed mostly unused dependency on findbugs.  In particular, the findbugs-core dependency on findbugs-annotations:2.0.0 will break Java 8 consumers.


elandau:
Any thoughts on using Guava's classpath scanner instead of asm?

governator-legacy/src/main/java/com/netflix/governator/lifecycle/AnnotationFinder.java
```diff
@@ -0,0 +1,206 @@
...
+    {
+        if(clazz == null)
+            clazz = classFromInternalName(className);
+        return clazz;
+    }
+
+    private Class<?> classFromInternalName(String name)
+    {
+        try {
+            return Class.forName(name.replace('/', '.'), true, classLoader);
```
elandau: Shouldn't the 2nd parameter be false so we don't force static initialization?

governator-legacy/src/main/java/com/netflix/governator/lifecycle/AnnotationFinder.java
```diff
@@ -0,0 +1,206 @@
...
+import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.objectweb.asm.Type.*;
+
+public class AnnotationFinder extends ClassVisitor
+{
+    private Set<Type> annotationTypes;
```
elandau: Make final

governator-legacy/src/main/java/com/netflix/governator/lifecycle/AnnotationFinder.java
```diff
@@ -0,0 +1,206 @@
...
+                                case BYTE: argClasses[i] = byte.class; break;
+                                case CHAR: argClasses[i] = char.class; break;
+                                case DOUBLE: argClasses[i] = double.class; break;
+                                case FLOAT: argClasses[i] = float.class; break;
+                                case INT: argClasses[i] = int.class; break;
+                                case LONG: argClasses[i] = long.class; break;
+                                case SHORT: argClasses[i] = short.class; break;
+                            }
+                        } catch(NullPointerException e) {
+                            System.out.println("here");
```
elandau: Log an error using slf4j if it's safe to ignore or throw an exception

governator-legacy/src/main/java/com/netflix/governator/lifecycle/ClasspathScanner.java
```diff
@@ -114,38 +113,66 @@ public ClasspathScanner(Collection<String> basePackages, Collection<Class<? exte
...
+                                    localClasses.addAll(finder.getAnnotatedClasses());
+                                    localMethods.addAll(finder.getAnnotatedMethods());
+                                    localConstructors.addAll(finder.getAnnotatedConstructors());
+                                    localFields.addAll(finder.getAnnotatedFields());
+                                }
+                            }
+                        }
+                        catch( IOException e )
+                        {
+                            throw new IllegalStateException(e);
```
elandau: Add jar name to exception to help debug errors

governator-legacy/src/main/java/com/netflix/governator/lifecycle/AnnotationFinder.java
jkschneider: In thought that as well, but was following what xbean-finder was doing internally.

governator-legacy/src/main/java/com/netflix/governator/lifecycle/AnnotationFinder.java
jkschneider: Sorry, this was a temporary debugging issue, will delete.

jkschneider:
Concerning Guava's classpath scanning utility, their docs say this:

> It's worth noting that ClassPath is a best-effort utility. It only scans classes in jar files or under a file system directory. Neither can it scan classes managed by custom class loaders that aren't URLClassLoader. **So don't use it for mission critical production tasks.**

governator-legacy/src/main/java/com/netflix/governator/lifecycle/ClasspathScanner.java
jkschneider: Thanks.  This also brought to my attention that I wasn't closing the JarFile resource.  Oops!

PR #231, opened 2015-08-11
Fix for classpathScanner and JAR files

mroloux:


elandau:
Thanks for the fix and PR.  I added this fix to #232 which has additional fixes.

PR #235, opened 2015-08-14
Fix bug testing current lifecycle state in LifecycleManager

elandau:


governator-core/src/main/java/com/netflix/governator/LifecycleManager.java
```diff
@@ -32,7 +32,7 @@
     public void addListener(LifecycleListener listener) {
         if (listeners.add(listener)) {
             LOG.info("Adding LifecycleListener '{}' {}", listener.getClass().getName(), System.identityHashCode(listener));
-            if (state.equals(State.Started)) {
+            if (state.get().equals(State.Started)) {
```
brharrington: I prefer using `==` for comparing enums because it will give you compile type type checking. For example, in this case `state == State.Started` would have failed to compile.

PR #241, opened 2015-09-18
Misc improvements

elandau:
Add template method initialize() to DefaultGovernatorConfiguration builder
Add feature to print all bindings immediately before the injector is created

### DIFF
--- a/governator-core/src/main/java/com/netflix/governator/Governator.java
+++ b/governator-core/src/main/java/com/netflix/governator/Governator.java
@@ -172,6 +172,12 @@ public class Governator {
                     createAutoModule(LOG, config, candidateModules, config.getModules()))
                    .with(config.getOverrideModules());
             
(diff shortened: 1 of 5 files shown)


governator-core/src/main/java/com/netflix/governator/Governator.java
```diff
@@ -172,6 +172,12 @@ public static LifecycleInjector createInjector(final GovernatorConfiguration con
                     createAutoModule(LOG, config, candidateModules, config.getModules()))
                    .with(config.getOverrideModules());
             
+            if (config.isEnabled(GovernatorFeatures.PRINT_BINDINGS)) {
```
brharrington: Why not just log it at debug level?

elandau: Good point.  I'll switch to that.

PR #245, opened 2015-10-01
Log class path scanning errors instead of failing.

elandau:


jhspaybar:
When does the scanning fail?  Is it when things like static dependencies don't resolve, or JNI type stuff doesn't find libraries, but the application didn't actually need those jars anyways to run?

LGTM though :+1:

PR #247, opened 2015-10-01
Fail when duplicate items found during class path scanning

elandau:


qiangdavidliu:
Related to #245, does these pulls essentially classify different governator failure modes where some are failable and some are just loggable? If so, it would be nice to have a clear guidance on which failure falls into which bucket.

PR #250, opened 2015-10-09
Deprecate AutoBindSingleton and log INFO messages where it's used.

elandau:
@AutoBindSingleton is a very misused feature of Governator and has caused numerous issues with applications having random classes annotated with @AutoBindSingleton being picked up and instantiated.  Furthermore, since @AutoBindSingleton is not a proper Guice Scope its use is misleading since @AutoBindSingleton classes injected without being picked up by the class path scanner aren't actually singletons.  

Moving forward it is recommended that any eager singleton should be specified in a Guice module and bound .asEagerSingleton().  

@AutoBindSingleton for modules is also being deprecated.  Instead, modules should be installed directly at injector creation or transitively via binder.install().

### DIFF
--- a/governator-annotations/src/main/java/com/netflix/governator/annotations/AutoBind.java
+++ b/governator-annotations/src/main/java/com/netflix/governator/annotations/AutoBind.java
@@ -27,7 +27,7 @@ import java.lang.annotation.Target;
  * AutoBindProvider to automatically/programmatically bind fields and constructor/method
  * arguments
  * 
- * @deprecated 10/10/2015 Deprecated in favor of standard Guice modules. 
+ * @deprecated 2015-10-10 Deprecated in favor of standard Guice modules. 
  *             See https://github.com/Netflix/governator/wiki/Auto-Binding
  */
 @Documented
--- a/governator-annotations/src/main/java/com/netflix/governator/annotations/AutoBind.java
+++ b/governator-annotations/src/main/java/com/netflix/governator/annotations/AutoBind.java
@@ -26,11 +26,14 @@ import java.lang.annotation.Target;
  * Annotation binding that combines with Governator's classpath scanning and a bound
  * AutoBindProvider to automatically/programmatically bind fields and constructor/method
  * arguments
+ * 
+ * @deprecated Deprecated in favor of standard Guice modules
  */
 @Documented
 @Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
 @Target({ElementType.FIELD, ElementType.CONSTRUCTOR, ElementType.METHOD, ElementType.ANNOTATION_TYPE, ElementType.PARAMETER})
 @Qualifier
+@Deprecated
 public @interface AutoBind
 {
     /**
--- a/governator-annotations/src/main/java/com/netflix/governator/annotations/AutoBind.java
+++ b/governator-annotations/src/main/java/com/netflix/governator/annotations/AutoBind.java
@@ -27,7 +27,8 @@ import java.lang.annotation.Target;
  * AutoBindProvider to automatically/programmatically bind fields and constructor/method
  * arguments
  * 
- * @deprecated Deprecated in favor of standard Guice modules
+ * @deprecated 10/10/2015 Deprecated in favor of standard Guice modules. 
+ *             See https://github.com/Netflix/governator/wiki/Auto-Binding
  */
 @Documented
 @Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
--- a/governator-annotations/src/main/java/com/netflix/governator/annotations/AutoBindSingleton.java
+++ b/governator-annotations/src/main/java/com/netflix/governator/annotations/AutoBindSingleton.java
@@ -22,7 +22,8 @@ import java.lang.annotation.Target;
 
 /**
  * Marks a class as a singleton. Governator will auto-bind it as an eager singleton
- * @deprecated  AutoBindSingleton is deprecated in favor of bindings in a Guice Module.
+ * @deprecated  10/10/2015 AutoBindSingleton is deprecated in favor of bindings in a Guice Module.
+ *              See https://github.com/Netflix/governator/wiki/Auto-Binding
  */
 @Documented
 @Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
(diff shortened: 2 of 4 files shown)


governator-annotations/src/main/java/com/netflix/governator/annotations/AutoBind.java
```diff
@@ -26,11 +26,14 @@
  * Annotation binding that combines with Governator's classpath scanning and a bound
  * AutoBindProvider to automatically/programmatically bind fields and constructor/method
  * arguments
+ * 
+ * @deprecated Deprecated in favor of standard Guice modules
```
jhspaybar: We might want to provide a wiki page with details on what to use instead.

governator-annotations/src/main/java/com/netflix/governator/annotations/AutoBindSingleton.java
```diff
@@ -22,10 +22,12 @@
 
 /**
  * Marks a class as a singleton. Governator will auto-bind it as an eager singleton
+ * @deprecated  AutoBindSingleton is deprecated in favor of bindings in a Guice Module.
```
jhspaybar: Same as above, a wiki might be nice that walks through the various uses and how to model them in the new way.

governator-legacy/src/main/java/com/netflix/governator/guice/InternalAutoBindModule.java
```diff
@@ -184,8 +186,14 @@ else if ( Module.class.isAssignableFrom(clazz) )
 
     private void bindAutoBindSingleton(AutoBindSingleton annotation, Class<?> clazz)
     {
-        LOG.info("Installing @AutoBindSingleton " + clazz.getName());
+        LOG.info("Installing @AutoBindSingleton '{}'", clazz.getName());
+        LOG.info("***** @AutoBindSingleton for '{}' is deprecated soon.\nPlease use a Guice module with bind({}.class).asEagerSingleton(); instead",
```
jhspaybar: This is definitely a better message than on the deprecated annotation, but we may still just want the wiki page.

howardyuan:
So this PR only logs the AutoBindSingleton, but not actually converts them into Singletons? How does this help Karyon3 users? I thought either we fail fast or we convert for them?

governator-legacy/src/main/java/com/netflix/governator/guice/InternalAutoBindModule.java
```diff
@@ -184,8 +186,14 @@ else if ( Module.class.isAssignableFrom(clazz) )
...
     private void bindAutoBindSingleton(AutoBindSingleton annotation, Class<?> clazz)
     {
-        LOG.info("Installing @AutoBindSingleton " + clazz.getName());
+        LOG.info("Installing @AutoBindSingleton '{}'", clazz.getName());
+        LOG.info("***** @AutoBindSingleton for '{}' is deprecated soon.\nPlease use a Guice module with bind({}.class).asEagerSingleton(); instead", 
+                clazz.getName(), clazz.getSimpleName() );
         
+        Singleton singletonAnnotation = clazz.getAnnotation(Singleton.class);
+        if (singletonAnnotation == null) {
+            LOG.info("***** {} should also be annotation with @Singleton to ensure singleton behavior", clazz.getName());
```
jhspaybar: annotation -> annotated

elandau: Converting to property singleton's is tricky and a horrible hack.  We should go through a proper deprecation cycle here.  Also, we shouldn't fail fast since this will break existing code.

governator-annotations/src/main/java/com/netflix/governator/annotations/AutoBind.java
elandau: I suppose we could just reference this github issue in code.

jhspaybar: I'd prefer we link to a page that lays out just the issue and fix for the user rather than having to parse a full thread to figure out what's relevant.

governator-legacy/src/main/java/com/netflix/governator/guice/InternalAutoBindModuleBootstrapModule.java
```diff
@@ -63,6 +63,7 @@ private Module bindAutoBindSingletons(BootstrapBinder binder) {
                         "@AutoBindSingleton(multiple=true) value cannot be set for Modules");
 
                 LOG.info("Found @AutoBindSingleton annotated module : {} ", clazz.getName());
+                LOG.info("***** @AutoBindSingleton for module {} is deprecated as of 10/10/2015.  Modules should be added directly to the injector or via install({}.class). See https://github.com/Netflix/governator/wiki/Auto-Binding", clazz.getName(), clazz.getSimpleName());
```
brharrington: I would tend to read that as saying something about that specific module is deprecated rather than AutoBindSingleton itself. Perhaps change the wording to something like:

> @AutoBindSingleton is deprecated as of Oct 10, 2015.  Module '{}' should be added directly to the injector or via install({}.class). See https://github.com/Netflix/governator/wiki/Auto-Binding

And as a minor change using a less [ambiguous date format](http://programmers.stackexchange.com/questions/42803/is-there-a-universal-date-format-that-anyone-in-the-world-can-understand).

For probable pit falls:

1. Since many internal users aren't using guice directly, it might be helpful to have the wiki cover the common ways to add modules for base-server, NFLifecycleUnitTester, etc. 
2. Is it assumed a given app owner will disable auto-binding as they switch? If so it would probably be good to add instructions to the wiki for how to do that. If not:
     * Do you think this should also check/warn about module deduping? Checking one of our modules that has AutoBindSingleton it doesn't currently implement equals/hashCode. If a user explicitly adds that module it will still get loaded via auto-binding for some time which will break with duplicate bindings.
     * Is there a way to detect and ignore auto-binding if the module is already bound? Otherwise the deprecation notice would keep showing up even though there is nothing to do.

elandau: Guice4 is actually A LOT better at handing module dedup without having to add hashCode() and equals() module is doing something bad like bind(Foo.class).toInstance(new FooImpl()). 

The deprecation notice only shows up if the module is found via classpath scanning.  The warning will no longer be logged once users have switched to installing the module directly and @AutoBindSingleton is removed.

brharrington: For the warning, the application owner that needs to explicitly install the module may be different than the library owner that needs to remove the AutoBindSingleton annotation. It seems like the primary consumer for the warning would be the application owner and as long as they are doing the right thing they probably shouldn't see the notice.

PR #252, opened 2015-11-25
Deprecate majority of governator-core which has moved to Karyon

elandau:


elandau:
We have decided to backport all the Karyon DI code back into governator instead.  This will be addressed in a separate pull request.

PR #256, opened 2015-12-11
WIP: ConditionalBinder

elandau:
The ConditionalBinder enables specifying multiple candidates for a binding with each having a condition that can be based on configuration or other state.  ConditionalBinder follows an implementation very similar to Multibinder in that all bindings are created with custom annotations and are resolved at injector creation time. 

For example,
```java
public class SnacksModule extends AbstractModule {
    protected void configure() {
      ConditionalBinder<Snack> conditionalbinder
          = ConditionalBinder.newConditionalBinder(binder(), Snack.class);
      multibinder.when(new ConditionalOnProperty("type", "twix")).toInstance(new Twix());
      multibinder.when(new ConditionalOnProperty("type", "snickers")).toProvider(SnickersProvider.class);
      multibinder.when(new ConditionalOnProperty("type", "skittles")).to(Skittles.class);
      multibinder.whenNone().to(Carrots.class);
     }
}
 ```

The following situations are not valid and result in a CreationException error at injector creation (not injection) time.  The exception will contain detailed information on the candidate bindings and where in code they were bound
#.  Multiple matching candidates
#.  No matched candidates
#.  Multiple default bindings specified
#.  Using conditional binding in Stage.PRODUCTION

Several approaches were considered for implementing this feature, including a Module annotation based approach.  The benefits of this approach are,
#.  Follows Guice conventions (such as Multibinder and MapBinder)
#.  Does not require Module post processing (which would complicate usage)
#.  Makes conditional processing injectable

Drawbacks to this approach are,
#.  Does not work in Stage.PRODUCTION (results in unmatched conditionals being instantiated).  Allowing this to run in production stage is possible but would add complexity that may not be necessary since we run only in the lazy Stage.DEVELOPMENT.  We can revisit this if running in Stage.PRODUCTION becomes are requirement.  
#.  Unmatched conditionals result in unused bindings

### DIFF
--- a/governator-core/src/main/java/com/netflix/governator/PropertiesPropertySource.java
+++ b/governator-core/src/main/java/com/netflix/governator/PropertiesPropertySource.java
@@ -4,19 +4,33 @@ import java.util.Properties;
 
 import javax.inject.Singleton;
 
+import com.google.inject.Binder;
 import com.google.inject.Module;
 import com.google.inject.Provides;
 import com.netflix.governator.spi.PropertySource;
 
-public class PropertiesPropertySource extends AbstractPropertySource {
+public final class PropertiesPropertySource extends AbstractPropertySource implements Module {
     private Properties props;
 
     public PropertiesPropertySource(Properties props) {
         this.props = props;
     }
 
+    public PropertiesPropertySource() {
+        this(new Properties());
+    }
+
     public static PropertiesPropertySource from(Properties props) {
         return new PropertiesPropertySource(props);
+    }
+    
+    public PropertiesPropertySource setProperty(String key, String value) {
+        props.setProperty(key, value);
+        return this;
+    }
+    
+    public boolean hasProperty(String key, String value) {
+        return props.containsKey(key);
     }
     
     public static Module toModule(final Properties props) {
@@ -38,4 +52,28 @@ public class PropertiesPropertySource extends AbstractPropertySource {
     public String get(String key, String defaultValue) {
         return props.getProperty(key, defaultValue);
     }
+
+    @Override
+    public void configure(Binder binder) {
+        binder.bind(PropertySource.class).toInstance(this);
+    }
+    
+    @Override
+    public int hashCode() {
+        return getClass().hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+
+        throw new RuntimeException("Only one PropertiesModule may be installed");
+    }
+
+
 }
--- a/governator-core/src/main/java/com/netflix/governator/conditional/Conditional.java
+++ b/governator-core/src/main/java/com/netflix/governator/conditional/Conditional.java
@@ -1,0 +1,15 @@
+package com.netflix.governator.conditional;
+
+
+/**
+ * Contract for any conditional that may be applied to a conditional binding
+ * bound via {@link ConditionalBinder}.
+ */
+public interface Conditional<T extends Conditional<T>> {
+    /**
+     * @return Class that can process this conditional.  This class
+     * is instantiated by Guice and is therefore injectable.  The class should be 
+     * annotated as a {@literal @}Singleton 
+     */
+    Class<? extends Matcher<T>> getMatcherClass();
+}
--- a/governator-core/src/main/java/com/netflix/governator/conditional/ConditionalBinder.java
+++ b/governator-core/src/main/java/com/netflix/governator/conditional/ConditionalBinder.java
@@ -1,0 +1,406 @@
+package com.netflix.governator.conditional;
+
+import java.lang.annotation.Annotation;
+import java.util.List;
+import java.util.Set;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+import com.google.inject.Binder;
+import com.google.inject.Binding;
+import com.google.inject.ConfigurationException;
+import com.google.inject.Inject;
+import com.google.inject.Injector;
+import com.google.inject.Key;
+import com.google.inject.Module;
+import com.google.inject.Provider;
+import com.google.inject.Stage;
+import com.google.inject.TypeLiteral;
+import com.google.inject.binder.LinkedBindingBuilder;
+import com.google.inject.internal.Errors;
+import com.google.inject.spi.BindingTargetVisitor;
+import com.google.inject.spi.Dependency;
+import com.google.inject.spi.Message;
+import com.google.inject.spi.ProviderInstanceBinding;
+import com.google.inject.spi.ProviderWithExtensionVisitor;
+import com.google.inject.spi.Toolable;
+
+/**
+ * An API to bind multiple conditional candidates separately, only to evaluate and 
+ * resolve to one candidate when the type is actually injected.  
+ * ConditionalBinder is intended for use in a Module
+ * 
+ * <pre><code>
+ * public class SnacksModule extends AbstractModule {
+ *   protected void configure() {
+ *     ConditionalBinder&lt;Snack&gt; conditionalbinder
+ *         = ConditionalBinder.newConditionalBinder(binder(), Snack.class);
+ *     multibinder.when(new ConditionalOnProperty("type", "twix")).toInstance(new Twix());
+ *     multibinder.when(new ConditionalOnProperty("type", "snickers")).toProvider(SnickersProvider.class);
+ *     multibinder.when(new ConditionalOnProperty("type", "skittles")).to(Skittles.class);
+ *     multibinder.whenNone().to(Carrots.class);
+ *   }
+ * }</code></pre>
+ *
+ * <p>With this binding when injecting {@code <Snack>} all conditionals will be evaluated and only
+ * the single matched binding will be injected
+ * <pre><code>
+ * class SnackMachine {
+ *   {@literal @}Inject
+ *   public SnackMachine(Snack snacks) { ... }
+ * }</code></pre>
+ *
+ * <p>Contributing conditional bindings from different modules is supported. For
+ * example, it is okay for both {@code CandyModule} and {@code ChipsModule}
+ * to create their own {@code ConditionalBinder<Snack>}, and to each contribute
+ * conditional bindings to the set of candidate snacks.  When a Snack is injected, it will 
+ * use the one conditional bindings that matched.
+ *
+ * Exactly one conditional binding may be matched.  An exception will be thrown 
+ * when Snack is injected and no or multiple bindings' conditions are matched.
+ * 
+ * <p>Conditionals are evaluated at injection time. If an element is bound to a
+ * provider, that provider's get method will be called each time Snack is
+ * injected (unless the binding is in Singleton scope, in which case Guice will cache
+ * the first call to get).
+ * 
+ * Conditionals may only be used in Stage.DEVELOPMENT since running in Stage.PRODUCTION will
+ * result in every Singleton conditional candidate being instantiated eagerly.  Any attempt
+ * to binding in Stage.PRODUCTION will result in a CreationException
+ * 
+ * TODO: Strip all conditional bindings of their scope so they cannot be eager singletons.
+ *       Alternatively, force lazy singleton behavior
+ * TODO: Discuss whether the conditional binder key should be singleton by default  
+ */
+public abstract class ConditionalBinder<T> {
+    /**
+     * Returns a new ConditionalBinder that tracks all candidate instances of {@code type}.
+     */
+    public static <T> ConditionalBinder<T> newConditionalBinder(Binder binder, TypeLiteral<T> type) {
+        return newRealConditionalBinder(binder, Key.get(type));
+    }
+
+    /**
+     * Returns a new ConditionalBinder that tracks all candidate instances of {@code type}.
+     */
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

File: governator-core/src/main/java/com/netflix/governator/conditional/ConditionalBinder.java [hunk 1]
```diff
@@ -0,0 +1,406 @@
...
+     * Each bound element must have a distinct value. Only the matching candidate will
+     * be instantiated and its provider cached when the key is first injected.
+     *
+     * <p>It is an error to call this method without also calling one of the
+     * {@code to} methods on the returned binding builder.
+     *
+     * <p>Scoping elements independently supported per binding. Use the {@code in} method
+     * to specify a binding scope.
+     */
+    public abstract LinkedBindingBuilder<T> whenMatch(Conditional obj);
```
tbak: What about adding operators to compose Conditionals (and, or/any, not)? Implementing this explicitly (by wrapping into higher level Conditional object) would be cumbersome, due to the way injections are handled.

elandau: Guice's Matcher has an abstract implementation that includes 'and' and 'or'.  We could follow the same pattern here.

elandau: I updated the PR with support for AND, OR and NOT operations on conditionals.  I also simplified the API by removing the Matcher class and making dependencies required for evaluating a conditional directly injectable into the conditional via member injection.

governator-core/src/main/java/com/netflix/governator/conditional/ConditionalBinder.java
```diff
@@ -0,0 +1,406 @@
...
+     * matching candidate will be instantiated and its provider cached when the key is 
+     * first injected.
+     *
+     * <p>It is an error to call this method without also calling one of the
+     * {@code to} methods on the returned binding builder.
+     *
+     * <p>Scoping elements independently supported per binding. Use the {@code in} method
+     * to specify a binding scope.
+     */
+    public abstract LinkedBindingBuilder<T> whenNoMatch();
```
jhspaybar: Can't we do this with whenMatch(new DefaultBindingConditional()) or something like that to avoid the special case?

governator-core/src/main/java/com/netflix/governator/conditional/Conditional.java
```diff
@@ -0,0 +1,37 @@
...
+ * member injection.
+ * 
+ */
+public abstract class Conditional {
+    /**
+     * Evaluate whether the condition is true.  evaluate() is only called once at injector
+     * creation time.
+     * @return True if conditional is true otherwise false
+     */
+    public abstract boolean evaluate();
```
jhspaybar: matches?

governator-core/src/main/java/com/netflix/governator/conditional/ConditionalBinder.java
```diff
@@ -0,0 +1,404 @@
...
+ * An API to bind multiple conditional candidates separately, only to evaluate and 
+ * resolve to one candidate when the type is actually injected.  
+ * ConditionalBinder is intended for use in a Module
+ * 
+ * <pre><code>
+ * public class SnacksModule extends AbstractModule {
+ *   protected void configure() {
+ *     ConditionalBinder&lt;Snack&gt; conditionalbinder
+ *         = ConditionalBinder.newConditionalBinder(binder(), Snack.class);
+ *     conditionalbinder.when(new ConditionalOnProperty("type", "twix")).toInstance(new Twix());
```
jhspaybar: should we allow this form?  feels like lots of room for screwing this up.

elandau: Which form are you referring to?  Do you mean the long DSL?  This is consistent with the Guice DSL, especially multibinder.

governator-core/src/main/java/com/netflix/governator/conditional/Conditional.java
elandau: I'm on the fence on what to call this.  Evaluate() seems more consistent with predicate terminology than matches().

governator-core/src/main/java/com/netflix/governator/conditional/ConditionalBinder.java
elandau: This seems like a very long and ambiguous way of specifying the default and is not in the sprit of a DSL.  DefaultBindingConditional could mean several things, is it the default implementation of a conditional or is it the conditional to use as the default.

jhspaybar: whenNone is just another conditional though, maybe call it whenMatch(new ConditionalOnNoOtherMatches())?

governator-core/src/main/java/com/netflix/governator/conditional/Conditional.java
jhspaybar: if(cond.evaluate()) seems much less clear to me than if(cond.matches()) but I don't feel strongly enough to push back if you want it this way.

governator-core/src/main/java/com/netflix/governator/conditional/Conditional.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Contract for any conditional that may be applied to a conditional binding
+ * bound via {@link ConditionalBinder}.
+ * 
+ * Dependencies needed by a concrete conditional should be injected using 
+ * member injection.
+ * 
+ */
+public abstract class Conditional {
```
jhspaybar: Sorry, I just noticed this is an abstract class rather than an interface.  Why?

governator-core/src/main/java/com/netflix/governator/conditional/ConditionalBinder.java
jhspaybar: Sorry, I was referring to the "toInstance()" form of binding since these are late bound it feels odd to instantiate something up front.

governator-core/src/main/java/com/netflix/governator/conditional/Conditional.java
elandau: Makes sense.  I'll change it back to matches().

governator-core/src/main/java/com/netflix/governator/conditional/Conditional.java
elandau: That's for the 'not' method.  This would not be an issue with interfaces in java8.

jhspaybar: Ah, could we extract the part that is the 3 interface methods then and turn Conditional into an enum with just the static method?

governator-core/src/main/java/com/netflix/governator/conditional/ConditionalBinder.java
elandau: Either way, ConditionalOnNoOtherMatches() would require special handling since it would require knowledge of how all other conditionals have been processed.  IMO it's much cleaner to just track this as an explicit special default case.

governator-core/src/main/java/com/netflix/governator/conditional/Conditional.java
elandau: Not sure I follow.  Can you provide a code example.

jhspaybar:
```
public enum Conditional {;
  public static Conditional not(Conditional conditional) {
    return new NotConditional(conditional);
  }
}

interface ConditionalMatcher {
  boolean matches(Injector injector);
  ConditionalMatcher and(ConditionalMatcher conditional);
  ConditionalMatcher or(ConditionalMatcher conditional);
}
```

Is what I had in mind.  It seems to me that and/or could also be static methods and pulled off onto the enum and turn the ConditionalMatcher into a single method interface for lambda goodness.  Your builder would then be like.
```
Conditional.not(Conditional.and(new ConditionA(), Conditional.or(new ConditionB(), new ConditionC())));
```

governator-core/src/main/java/com/netflix/governator/conditional/ConditionalBinder.java
jhspaybar: How do I do something like whenMatch(Conditional.and(new ConditionOnNoOtherMatches(), new ConditionalOnProfile("prod")) if this is kept as a special case?

governator-core/src/main/java/com/netflix/governator/conditional/Conditional.java
```diff
@@ -0,0 +1,20 @@
...
+ * Contract for any conditional that may be applied to a conditional binding
+ * bound via {@link ConditionalBinder}.
+ * 
+ * Dependencies needed by a concrete conditional should be injected using 
+ * member injection.
+ * 
+ */
+public interface Conditional {
+    /**
+     * Evaluate whether the condition is true.  evaluate() is only called once at injector
```
qiangdavidliu: javadoc update

governator-core/src/main/java/com/netflix/governator/PropertiesPropertySource.java
```diff
@@ -4,21 +4,35 @@
 
 import javax.inject.Singleton;
 
+import com.google.inject.Binder;
 import com.google.inject.Module;
 import com.google.inject.Provides;
 import com.netflix.governator.spi.PropertySource;
 
-public class PropertiesPropertySource extends AbstractPropertySource {
+public final class PropertiesPropertySource extends AbstractPropertySource implements Module {
```
tcellucci: Gak.  Properties is legacy Java with all synchronized methods.  Why do we want this in 2016, and not a Map<String,String>?

governator-core/src/main/java/com/netflix/governator/PropertiesPropertySource.java
```diff
@@ -38,4 +52,28 @@ public String get(String key) {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+
+        throw new RuntimeException("Only one PropertiesModule may be installed");
```
tcellucci: IllegalStateException would be clearer; usually a bad idea to throw RuntimeException

governator-core/src/main/java/com/netflix/governator/conditional/AllOfConditional.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.inject.Injector;
+
+/**
+ * Conditional that is true if and only if all child conditionals are true
+ */
+public class AllOfConditional implements Conditional {
+    private final List<Conditional> children;
+    
+    public AllOfConditional(List<Conditional> children) {
```
tcellucci: should be able to take any Iterable or Collection as input, and add a varargs constructor too?

governator-core/src/main/java/com/netflix/governator/conditional/ConditionalBinder.java
```diff
@@ -0,0 +1,409 @@
+package com.netflix.governator.conditional;
+
+import java.lang.annotation.Annotation;
+import java.util.List;
+import java.util.Set;
+
+import com.google.common.collect.ImmutableList;
```
tcellucci: do we need to use these google collections classes any more (with Java 8)?

governator-core/src/main/java/com/netflix/governator/conditional/ConditionalBinder.java
```diff
@@ -0,0 +1,409 @@
...
+     * the qualifier {@code annotation}
+     */
+    public static <T> ConditionalBinder<T> newConditionalBinder(Binder binder, Class<T> type,
+            Class<? extends Annotation> annotationType) {
+        return newRealConditionalBinder(binder, Key.get(type, annotationType));
+    }
+
+    static <T> ConditionalBinder<T> newRealConditionalBinder(Binder binder, Key<T> key) {
+        if (binder.currentStage().equals(Stage.PRODUCTION)) {
+            throw new RuntimeException("ConditionalBinder may not be used in Stage.PRODUCTION.  Use Stage.DEVELOPMENT.");
```
tcellucci: IllegalStateException

governator-core/src/main/java/com/netflix/governator/conditional/ConditionalElementImpl.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "@" + ConditionalElement.class.getName() + "(keyName=" + keyName
+                + ",uniqueId=" + uniqueId + ",condition=" + condition + ")";
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
tcellucci: seems flawed that Condition is not part of the equality check, but is part of the object state and passed in via constructor.  I would include at least an identity comparison for Condition here.

governator-core/src/test/java/com/netflix/governator/conditional/ConditionalBinderTest.java
```diff
@@ -0,0 +1,346 @@
...
+        
+        @Override
+        public String getName() {
+            return name;
+        }
+    }
+    
+    @Singleton
+    static class SingletonFoo implements Foo {
+        static int injectedCount = 0;
```
tcellucci: volatile?

tcellucci:
I didn't see any unit tests that use the @ConditionalElement; I think we need tests to show these are processed correctly prior to merge.  Annotation-based conditionals are supported yes?

PR #274, opened 2016-02-15
removed redundant and outdated servlet-api dependency

twicksell:
javax.servlet-api is already coming from jetty-server. This dependency was outdated and causing multiple  versions of the servlet-api to be present on the classpath.


drobertduke:
:+1:

qiangdavidliu:
:+1:

PR #276, opened 2016-02-19
Fix @PostConstruct and @PreDestory for @Provides methods

elandau:
Use ProvisionListener instead of TypeListener so @PreDestroy and @PostConstruct on object returned by @Provides methods are called.


governator-core/src/main/java/com/netflix/governator/LifecycleModule.java
```diff
@@ -144,62 +141,59 @@ public String toString() {
...
-                                        } 
-                                        catch (Exception e) {
-                                            LOG.error("Failed to call @PreDestroy method {} on {}", new Object[]{m, injectee.getClass().getName()}, e);
-                                        }
-                                    }
+                if (injectee instanceof LifecycleListener) {
+                    pendingLifecycleListeners.add((LifecycleListener)injectee);
+                }
+                
+                // TODO: Add to PreDestroy list
```
twicksell: Still applicable?

PR #277, opened 2016-02-26
Enhancements to Jersey-Guice integration

elandau:
*  Configure Jersey via property bindings (instead of statically in JerseyServletModule)
*  Auto create all singleton resources

To enable just add the following binding

```java
bind(GuiceContainer.class).to(GovernatorServletContainer.class).asEagerSingleton();
```

To set up a Jersey endpoint,
```java
serve("/*").with(GuiceContainer.class);  
```

To customize the configuration of the Jersey endpoint
 
```java
@Provides
ResourceConfig getResourceConfig() {
    return new PackagesResourceConfig(ImmutableMap.<String, Object>builder()
            .put(PackagesResourceConfig.PROPERTY_PACKAGES, "org.example.resources")
            .put(ResourceConfig.FEATURE_DISABLE_WADL, "false")
            .build());
}
```


governator-jersey/src/main/java/com/netflix/governator/guice/jetty/GovernatorJerseyServletModule.java
```diff
@@ -0,0 +1,55 @@
...
+ * by setting the property, com.sun.jersey.config.feature.DisableWADL=true.  Also, when running
+ * in Stage.DEVELOPMENT all scanning singleton resources will be eagerly created.
+ * 
+ */
+public abstract class GovernatorJerseyServletModule extends JerseyServletModule {
+    private static final class InternalGovernatorJerseyServletModule extends ServletModule {
+        @Override
+        protected void configureServlets() {
+            bind(GuiceContainer.class).to(GovernatorServletContainer.class).asEagerSingleton();
+            serve("/*").with(GuiceContainer.class);
```
twicksell: Should we consider making the prefix configurable?

governator-jersey/build.gradle
```diff
@@ -0,0 +1,16 @@
+apply plugin: 'java'
+
+dependencies {
+    compile project(':governator-core')
+
+    compile 'javax.servlet:servlet-api:2.5'
```
twicksell: Don't make this a compile dependency. It should be providedRuntime. Also consider upgrading to a more modern version of the servlet api.

governator-jersey/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+
+    compile 'com.google.inject.extensions:guice-servlet:4.0'
+
+    compile 'com.sun.jersey:jersey-server:1.19'
+    compile 'com.sun.jersey.contribs:jersey-guice:1.19'
+
+    compile 'org.eclipse.jetty:jetty-servlet:9.2.12.v20150709'
```
twicksell: This also seems like a sketchy dependency for compile scope

governator-jersey/src/main/java/com/netflix/governator/jersey/annotations/ServletContainerProperties.java
```diff
@@ -0,0 +1,15 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import javax.inject.Qualifier;
+
+@Qualifier
+@Target({ElementType.FIELD, ElementType.PARAMETER, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ServletContainerProperties {
+
```
twicksell: What is the purpose of this annotation?

governator-jersey/src/main/java/com/netflix/governator/guice/jetty/GovernatorJerseyServletModule.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Enhancement to JerseyServletModule by allowing the GuiceContainer and Jersey to be configured
+ * via bindings instead of statically in a Module.
+ * 
+ * You can provide the configuration values to Jersey using the binding
+ * 
+ * ```java
+ *  @Provides
+ *  @ServletContainerProperties 
+ *  Map<String, Object> getJerseyProperties() {
```
twicksell: How do I specify different properties from separate modules without overwriting the entire map? Would it not be better to get these from archaius?

settings.gradle
```diff
@@ -7,5 +7,6 @@ include 'governator-commons-cli'
 include 'governator-test-junit'
 include 'governator-archaius'
 include 'governator-jetty'
+include 'governator-jersey'
```
twicksell: jersey1? Not sure if jersey2 actually requires us to implement anything similar at this point.

governator-jersey/src/main/java/com/netflix/governator/guice/jetty/GovernatorJerseyServletModule.java
elandau: Couldn't figure out how to do that yet :)

governator-jersey/build.gradle
elandau: ok

governator-jersey/src/main/java/com/netflix/governator/jersey/annotations/ServletContainerProperties.java
elandau: I'm gonna get rid of this

governator-jersey/src/main/java/com/netflix/governator/guice/jetty/GovernatorJerseyServletModule.java
elandau: I'll request a binding to ResourceConfig instead of this so we can tie it to archaius.

settings.gradle
elandau: jersey2 should be different.  I was planning on governator-jersey and governator-jersey2

governator-jersey/src/main/java/com/netflix/governator/guice/jetty/GovernatorJerseyServletModule.java
```diff
@@ -0,0 +1,55 @@
...
+ * In addition installing this module will automatically turn on the WADL (which can be turned off
+ * by setting the property, com.sun.jersey.config.feature.DisableWADL=true.  Also, when running
+ * in Stage.DEVELOPMENT all scanning singleton resources will be eagerly created.
+ * 
+ */
+public abstract class GovernatorJerseyServletModule extends JerseyServletModule {
+    private static final class InternalGovernatorJerseyServletModule extends ServletModule {
+        @Override
+        protected void configureServlets() {
+            bind(GuiceContainer.class).to(GovernatorServletContainer.class).asEagerSingleton();
```
qiangdavidliu: The API here is a little different from what JerseyServletModule requires (the servletModule requires explicit serving of GuiceContainer in the configureServlets, whereas this module implicitly does this). It would be nice to be consistent with JerseyServletModule.

governator-jersey/src/main/java/com/netflix/governator/guice/jetty/GovernatorServletContainer.java
```diff
@@ -0,0 +1,92 @@
...
+ *     return new PackagesResourceConfig(ImmutableMap.<String, Object>builder()
+ *             .put(PackagesResourceConfig.PROPERTY_PACKAGES, "org.example.resources")
+ *             .put(ResourceConfig.FEATURE_DISABLE_WADL, "false")
+ *             .build());
+ * }
+ * ```
+ * 
+ * In addition installing this module will automatically turn on the WADL (which can be turned off
+ * by setting the property, com.sun.jersey.config.feature.DisableWADL=true.  Also, when running
+ * in Stage.DEVELOPMENT all scanning singleton resources will be eagerly created.
```
qiangdavidliu: This point is a major behavioural change, so would be good to document this new behaviour at the top of the comments block instead of as an afterthought here.

PR #278, opened 2016-03-01
Fix Servlet Context Listener shutdown bug.

drtechniko:
The GovernatorServletContextListener never set its internal
injector instance so at shutdown it never sent a signal to the lifecycle
listener to call all the @PreDestroy methods (among things) when the
application shuts down and before the JVM does.

### DIFF
--- a/governator-servlet/src/main/java/com/netflix/governator/guice/servlet/GovernatorServletContextListener.java
+++ b/governator-servlet/src/main/java/com/netflix/governator/guice/servlet/GovernatorServletContextListener.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


drtechniko:
@elandau @twicksell Please have a look. This fixes my issue with the app not shutting down properly.

governator-servlet/src/main/java/com/netflix/governator/guice/servlet/GovernatorServletContextListener.java
```diff
@@ -85,12 +99,15 @@ public void contextDestroyed(ServletContextEvent servletContextEvent) {
      */
     protected final Injector getInjector() {
         try {
-            return createInjector();
+            if (injector == null) {
```
elandau: We should throw an exception if injector != null

elandau:
nice catch!

governator-servlet/src/main/java/com/netflix/governator/guice/servlet/GovernatorServletContextListener.java
```diff
@@ -82,15 +96,21 @@ public void contextDestroyed(ServletContextEvent servletContextEvent) {
     /**
      * Override this method to create (or otherwise obtain a reference to) your
      * injector.
+     * NOTE: If everything is set up right, then this method should only be called once during
+     * application startup.
      */
     protected final Injector getInjector() {
+        if (injector != null) {
+            throw new RuntimeException("We must only create an Injector once.");
```
elandau: Actually, this should be an IllegalStateException.  Wording should be 'Injector already created'.

PR #280, opened 2016-03-08
add nullchecks to provision listeners and previously failing test case

twicksell:


governator-core/src/test/java/com/netflix/governator/NullableBindingTest.java
```diff
@@ -0,0 +1,61 @@
...
+public class NullableBindingTest {
+
+    private Injector injector;
+
+    @Before
+    public void setup() {
+        injector = InjectorBuilder.fromModules(new GuavaApplicationEventModule(), new AbstractModule() {
+            @Override
+            protected void configure() {
+                bind(InnerDependency.class).toProvider(new Provider<InnerDependency>() {
```
twicksell: FWIW an instance binding will still throw an exception during injector creation. But Providers do support Nullable.

elandau: Use Providers.of(null)

PR #282, opened 2016-03-30
JDiff report generation and publication enabled.

noel-yap:


build.gradle
```diff
@@ -49,3 +51,90 @@ subprojects {
...
+/**
+ * Generate a JDiff report between a newer version and an older version.
+ *
+ * Usage:
+ *  gradle -PjDiff.olderVersion=2.1312.0 [-PjDiff.newerVersion=2.1716.0] generateJDiffReport
+ *
+ * View generated report at:
+ *  build/reports/jDiff/${olderVersion}-to-${newerVersion}/changes.html
+ */
+task generateJDiffReport(dependsOn: '_jDiff_cloneOldVersion') {
```
twicksell: Why not just package this task inside of a nebula plugin? Seems like a better home

noel-yap: Creating a full-fledged plugin will take more time. I want to get some usage while that work happens so as to provide input as to how make the plugin more easily usable.

PR #291, opened 2016-05-03
Install LifecycleModule first so @PostConstruct support is added early

elandau:


governator-core/src/main/java/com/netflix/governator/LifecycleInjectorCreator.java
```diff
@@ -93,7 +93,6 @@ public LifecycleInjector createInjector(Stage stage, Module module) {
             onBeforeInjectorCreate();
             Injector injector = Guice.createInjector(
                 stage, 
-                module,
                 new LifecycleModule(),
```
qiangdavidliu: Be good to have some comments here about ordering matter here and the LifecycleModule must be the first Module installed.

howardyuan:
LGTM

qiangdavidliu:
👍

PR #292, opened 2016-05-06
Move JDiff-related code to jdiff.gradle since it's causing issues wit…

noel-yap:
…h IDEs.


elandau:
Do the JDiff reports still run?

noel-yap:
@elandau , not by default. The build will need to use `-b jdiff.gradle` in order to produce the JDiff reports.

PR #293, opened 2016-05-06
backport new governator LifecycleListener support to legacy injectorbuilder

twicksell:


governator-core/src/main/java/com/netflix/governator/LifecycleListenerModule.java
```diff
@@ -0,0 +1,86 @@
...
+                if (manager == null) {
+                    pendingLifecycleListeners.add((LifecycleListener) injectee);
+                } else {
+                    manager.addListener((LifecycleListener) injectee);
+                }
+            }
+        }
+    }
+
+    @Override
```
elandau: This module has state so let's not dedup

governator-core/src/main/java/com/netflix/governator/LifecycleListenerModule.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.inject.AbstractModule;
+import com.google.inject.matcher.Matchers;
+import com.google.inject.spi.ProvisionListener;
+import com.netflix.governator.annotations.SuppressLifecycleUninitialized;
+import com.netflix.governator.spi.LifecycleListener;
+
+/**
+ * Adds support for detection and invocation of {@link LifecycleListener} instances. 
+ */
+public class LifecycleListenerModule extends AbstractModule {
```
elandau: Make this 'final'

governator-legacy/build.gradle
```diff
@@ -8,4 +8,6 @@ dependencies {
     compile    'com.fasterxml.jackson.core:jackson-databind:2.4.3'
 
     testCompile 'org.hamcrest:hamcrest-library:1.3'
+    testCompile 'junit:junit:latest.release'
```
elandau: Does latest.release work in oss?

governator-legacy/src/main/java/com/netflix/governator/lifecycle/LifecycleManagerArguments.java
```diff
@@ -146,4 +150,12 @@ public void setConfigurationDocumentation(ConfigurationDocumentation configurati
     public ConfigurationDocumentation getConfigurationDocumentation() {
         return configurationDocumentation;
     }
+
+    public LifecycleManager getLifecycleManager() {
```
elandau: Can we make this package private?

governator-legacy/src/test/java/com/netflix/governator/lifecycle/LifeCycleFeaturesOnLegacyBuilderTest.java
```diff
@@ -0,0 +1,69 @@
...
+import com.netflix.governator.guice.LifecycleInjector;
+import com.netflix.governator.spi.LifecycleListener;
+
+@RunWith(MockitoJUnitRunner.class)
+public class LifeCycleFeaturesOnLegacyBuilderTest {
+    
+    private com.netflix.governator.lifecycle.LifecycleManager legacyLifecycleManager;
+
+    @Spy
+    private final TestListener listener = new TestListener();
```
elandau: Why spy and not mock?

twicksell: I didn't care about the behavior of this class since all of the methods are no-ops. So a spy seemed logical.

governator-legacy/build.gradle
twicksell: It does. I'll make these a specific version so its consistent, but given that they are testCompile and not transitive I don't believe it makes much of a difference.

PR #296, opened 2016-05-19
Deprecate warmup

elandau:
@WarmUp is somewhat of an anti-pattern as it allow partially formed objects to be injected.  This removes all @WarmUp code but keeps the annotation and treats @WarmUp the same way as @PostConstruct.


twicksell:
LGTM

PR #297, opened 2016-05-25
first pass at AutoCloseable support and enhancements for PostConstruc…

tcellucci:
- add AutoCloseable support to PreDestroyLifecycleActions
- improve JSR-250 compliance for PostConstruct and PreDestroy processing
- implement AutoCloseable in LifecycleInjector (enable try / resource construct with LifecycleInjector)
- add LifecycleInjectorBuilder.fromInstance() to reduce boilerplate code with AbstractModule {}
- add guard conditions for class introspection by PostConstruct/PreDestroyLifecycleActions


governator-core/src/main/java/com/netflix/governator/InjectorBuilder.java
```diff
@@ -62,6 +63,23 @@ public static InjectorBuilder fromModules(List<Module> modules) {
         return new InjectorBuilder(Modules.combine(modules));
     }
     
+    public static <T> InjectorBuilder fromInstance(final Object o) {
+        return fromInstances(o);
+    }
+    
+    public static InjectorBuilder fromInstances(final Object... instances) {
```
twicksell: If we do this, maybe consider renaming the method to something like "bindInstances".

governator-core/src/main/java/com/netflix/governator/InjectorBuilder.java
```diff
@@ -62,6 +63,23 @@ public static InjectorBuilder fromModules(List<Module> modules) {
         return new InjectorBuilder(Modules.combine(modules));
     }
     
+    public static <T> InjectorBuilder bindInstance(final Object o) {
```
elandau: We probably want to avoid this since we're trying to stay true to the Guice API.  We can discuss this further but either way let's keep this separate from this PR.

governator-core/src/main/java/com/netflix/governator/internal/AbstractLifecycleFeature.java
```diff
@@ -17,39 +20,67 @@
...
+abstract class AbstractLifecycleFeature<C> implements LifecycleFeature {
+	enum IntrospectionFlag {
+		FIELD, METHOD, INTERFACE, SUPERCLASS
+	}
 
-    private void visitFieldsAndMethods(Class<?> type, List<LifecycleAction> actions) {
+	private Set<IntrospectionFlag> flags;
+	
+	private AbstractLifecycleFeature(Set<IntrospectionFlag> flags) {
+		this.flags = flags;
```
elandau: Indentation looks funky.  We've been using 4 space indentation in Governator.

governator-core/src/test/java/com/netflix/governator/PreDestroyTest.java
```diff
@@ -0,0 +1,240 @@
...
+		try (LifecycleInjector injector = builder.createInjector()) {
+			Assert.assertNotNull(injector.getInstance(Foo.class));
+			Mockito.verify(foo, Mockito.never()).shutdown();
+		}        
+        Mockito.verify(foo, Mockito.times(1)).shutdown();
+    }
+    
+ 
+    
+    @Before
```
elandau: Can we create a separate JUnit rule for this?  Or do this via a runner?

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyLifecycleActions.java
```diff
@@ -2,55 +2,103 @@
...
- * Note that this feature is implicit in LifecycleModule and therefore does not need 
- * to be added using the LifecycleFeature multibinding.
+ * Special AbstractLifecycleFeature to support @PreDestroy annotation processing and
+ * java.io.AutoCloseable detection. Note that this feature is implicit in LifecycleModule 
+ * and therefore does not need to be added using the LifecycleFeature multibinding.
  * 
  * @author elandau
  */
-public final class PreDestroyLifecycleActions extends AbstractLifecycleFeature {
+public final class PreDestroyLifecycleActions extends AbstractLifecycleFeature<Set<String>> {
```
elandau: Do we really need a generic list here?  Seems like this is only relevant for methods.  We can just use a Set<Method> that is passed to getMethodActions()

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyLifecycleActions.java
```diff
@@ -2,55 +2,103 @@
...
+	        typeActions = super.getActionsForType(type);
+		}
+		return typeActions;
+	}
+
+	@Override
+    protected List<LifecycleAction> getMethodActions(Set<String> typeContext, final Class<?> type, final Method method) {
+		List<LifecycleAction> typeActions=null;
+    	int modifiers = method.getModifiers();
+		if (!Modifier.isStatic(modifiers) && !Modifier.isAbstract(modifiers) && method.getParameterCount() == 0  && Void.TYPE.equals(method.getReturnType())) {
```
elandau: We should probably throw an exception if someone is annotating a static method with @PreDestroy or @PostConstruct as it won't work as they would be expecting.

governator-core/src/main/java/com/netflix/governator/internal/AbstractLifecycleFeature.java
```diff
@@ -17,39 +20,67 @@
  * 
  * @author elandau
  */
-abstract class AbstractLifecycleFeature implements LifecycleFeature {
+abstract class AbstractLifecycleFeature<C> implements LifecycleFeature {
+	enum IntrospectionFlag {
+		FIELD, METHOD, INTERFACE, SUPERCLASS
```
elandau: Under what circumstances would INTERFACE or SUPERCLASS be optional?

governator-core/src/main/java/com/netflix/governator/InjectorBuilder.java
topcat-70: that InjectorBuilder method reduces the test class size by about half for this commit.   It saves boilerplate construction of an inner class extending AbstractModule and binding a single instance by type, with every test method.  Given that the purpose of a 'Builder' pattern like InjectorBuilder is to reduce boilerplate construction for the Injector, I think there's merit to a convenience method like  the bindInstance() methods

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyLifecycleActions.java
topcat-70: AbstractLifecycleFeature has 4 introspection methods of any consequence, and the new 'context' argument needs to be a consistently-typed parameter passed into each one.  Using a generic signature for the context type means no casting for the subclasses, and flexibility for other implementations to use java.util.Map or whatever type works best for them (possibly null).  Coordination of argument types across methods seems legitimate use of generics; is there a down side?

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyLifecycleActions.java
topcat-70: y, let's talk about how we want to handle failures; there are several new modes added with this commit.

governator-core/src/main/java/com/netflix/governator/internal/AbstractLifecycleFeature.java
topcat-70: INTERFACE would be optional if we want JSR-250-recommended handling of @PreDestroy and @PostConstruct.  SUPERCLASS would be optional if we only wanted to consider annotations in the top-level class. for instance we might process @Cacheable or @Bean only at the top-level to 'minimize astonishment' for the coder.  Practically I added SUPERCLASS to the enum for consistency's sake.  It would otherwise be the only one of four introspection control points with no 'feature switch', and the cost to add it was vanishingly small.

governator-core/src/test/java/com/netflix/governator/PreDestroyTest.java
topcat-70: Let's talk about this one; not sure what you mean

governator-core/src/main/java/com/netflix/governator/internal/AbstractLifecycleFeature.java
topcat-70: will fix.  is there a style guide somewhere, or do we use someone's standard?

governator-core/src/main/java/com/netflix/governator/InjectorBuilder.java
elandau: The InjectorBuilder currently deals with modules as opposed to introducing a new binding API.  I think this needs a lot more discussion before we expand the scope of the InjectorBuilder.   Let's do this as a separate pull request.

governator-core/src/main/java/com/netflix/governator/internal/AbstractLifecycleFeature.java
elandau: There is not a styleguide but we should definitely add one.

governator-core/src/main/java/com/netflix/governator/InjectorBuilder.java
elandau: Instead of complicating the InjectorBuilder, this syntactic sugar can better be addressed by moving this into a separate class. 

```java
public static final ModuleUtils {
    public static Module fromInstances(final Object... instances) {
        return new AbstractModule() {
            @SuppressWarnings("unchecked")
            @Override
            protected void configure() {
             	for (Object o : instances) {
             		bind((Class)o.getClass()).toInstance(o);
             	}
         };
    }
}
```

Your unit tests would then look like, 
```java
InjectorBuilder.fromModules(ModuleUtils.fromInstances(foo)).createInjector()
```

governator-core/src/main/java/com/netflix/governator/internal/PostConstructLifecycleActions.java
```diff
@@ -16,39 +19,59 @@
...
-                        .toString();
-                }
-            });
-        }
+    protected List<LifecycleAction> getMethodActions(Set<String> typeContext, final Class<?> type, final Method method) {
+    	int modifiers = method.getModifiers();
+		if (!Modifier.isStatic(modifiers) && !Modifier.isAbstract(modifiers) && method.getParameterCount() == 0  && Void.TYPE.equals(method.getReturnType())) {
+			if (!typeContext.contains(method.getName())) {
+		        if (null != method.getAnnotation(PostConstruct.class)) {
+			        if (!method.isAccessible()) {
```
elandau: Is this check necessary?  Ultimately we always need to make the method accessible.

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyLifecycleActions.java
elandau: This seems like too much 'flexibility' and would result in a confusing API.  Why not just make so that getMethodActions takes a Set<Method> and getFieldActions task a Set<Field>?  Actually, what would be the use case for field actions?

topcat-70: I think forcing all subclasses to work with a 'Set<String>' type for the context, when we can't know exactly how they'll be using the context, is needlessly restrictive with only slight reduction in complexity.  If subclasses are forced to receive a 'Set' or 'Set<?>', then they may as well get a java.lang.Object, as the amount of casting required will be the same to access its content.

governator-core/src/main/java/com/netflix/governator/internal/PostConstructLifecycleActions.java
topcat-70: no need to make a Method accessible if it already is; I'd expect .isAccessible() is cheaper call than .setAccesible() even in the trivial case.

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyLifecycleActions.java
topcat-70: refactored this code to use a visitor pattern, so the generic parameter isn't needed

governator-core/src/main/java/com/netflix/governator/internal/AbstractLifecycleFeature.java
topcat-70: code reformatted with 4-space indentation

governator-core/src/main/java/com/netflix/governator/InjectorBuilder.java
topcat-70: moved logic out of InjectorBuilder; now looks like this:

        try (LifecycleInjector injector = TestSupport.inject(postConstructChild)) {
           [ ...]
        }

governator-core/src/main/java/com/netflix/governator/LifecycleInjector.java
```diff
@@ -102,4 +102,9 @@ public void shutdown() {
     public void addListener(LifecycleListener listener) {
         manager.addListener(listener);
     }
+    
+    @Override
+    public void close() {
+    	shutdown();
```
elandau: Should we deprecated shutdown() and just have close()?

governator-core/src/main/java/com/netflix/governator/internal/AbstractLifecycleFeature.java
```diff
@@ -1,55 +1,71 @@
...
- * getFieldAction and getMethodAction.
+ * Base utility class for field and method based lifecycle features. This class
+ * deals with identifying all the fields and methods for the type and its
+ * superclasses and interfaces and delegating the actual processing using the
+ * template methods getFieldAction and getMethodAction.
  * 
  * @author elandau
  */
 abstract class AbstractLifecycleFeature implements LifecycleFeature {
+    interface TypeVisitor {
```
elandau: I like the idea of using a visitor here but this visitor doesn't exactly conform to the standard pattern.  

How about a basic visitor like this,

```java
public static class VisitableType {
    public interface Visitor {
        void visitMethod(Method method);
        void visitField(Field field);
        void visitSuper(Class<?> type);
        void visitInterface(Class<?> type);
    }
    
    private final Class<?> type;
    
    public VisitableType(Class<?> type) {
        this.type = type;
    }
    
    <T extends Visitor> T accept(T visitor) {
        if (type == null) 
            return visitor;
        
        for (final Field field : type.getDeclaredFields()) {
            visitor.visitField(field);
        }
 
        for (final Method method : type.getDeclaredMethods()) {
            visitor.visitMethod(method);
        }
 
        visitor.visitSuper(type.getSuperclass());

        for (Class<?> iface : type.getInterfaces()) {
            visitor.visitInterface(iface);
        }
        
        return visitor;
    }
}
```

The PostConstructVisitor would look something like this,

```java
public static class PostConstructVisitor implements VisitableType.Visitor {
    private List<LifecycleAction> postConstruct = new ArrayList<>();
    private final boolean includeSupersAndInterfaces;

    public PostConstructVisitor(boolean includeSupersAndInterfaces) {
         this.includeSupersAndInterfaces = includeSupersAndInterfaces;
    }

    @Override
    public void visitMethod(Method method) {
        // For brevity I'm omitting all the additional processing and checks
        if (method.getAnnotation(PostConstruct.class) != null) {
            postConstruct.add(new LifecycleAction() {
                @Override
                public void call(Object obj) throws Exception {
                    method.invoke(obj);
                }
            });
        }
    }

    @Override
    public void visitField(Field field) {
    }

    @Override
    public void visitSuper(Class<?> type) {
        if (includeSupersAndInterfaces && !hasPostConstruct()) {
            new VisitableType(type).accept(this);
        }
    }

    @Override
    public void visitInterface(Class<?> type) {
        if (includeSupersAndInterfaces && !hasPostConstruct()) {
            new VisitableType(type).accept(this);
        }
    }
    
    private boolean hasPostConstruct() {
        return !postConstruct.isEmpty();
    }
    
    public List<LifecycleAction> getActions() {
        return postConstruct;
    }
}
```

Usage is simply this,

```java
List<LifecycleActions> actions = new VisitableType(SomeType.class).accept(new PostConstructVisitor()).getActions();
```

You can then further customize PostConstructVisitor to adhere to either the existing behavior or that of the JSR.

Note that even though there may be additional overhead for iterating all elements of the class that's fine because the results are cached when the type is first encountered.

governator-core/src/main/java/com/netflix/governator/LifecycleInjector.java
topcat-70: we can deprecate it, but I'd definitely leave shutdown() present to avoid a breaking change

topcat-70:
bogus slf4j issue; works fine locally.  Evidently I can clear things up with adding a "new Object[]{ ...} around the method arguments

governator-core/src/main/java/com/netflix/governator/internal/PostConstructLifecycleActions.java
```diff
@@ -1,57 +1,119 @@
...
+        private PostConstructAction(Method method) {
+            this.method = method;
+            this.description = new StringBuilder().append("PostConstruct[").append(method.getDeclaringClass().getName())
+                    .append("#").append(method.getName()).append("]").toString();
+        }
+
+        @Override
+        public void call(Object obj)
+                throws IllegalAccessException, IllegalArgumentException, InvocationTargetException {
+            LOG.info("invoking lifecycle action {}", description);
```
elandau: info might be too noisy.  Let's make this debug.

governator-core/src/main/java/com/netflix/governator/internal/PostConstructLifecycleActions.java
```diff
@@ -1,57 +1,119 @@
...
+            return !clazz.isInterface();
+        }
+
+        @Override
+        public boolean visit(final Method method) {
+            int modifiers = method.getModifiers();
+            if (method.isAnnotationPresent(PostConstruct.class)) {
+                String methodName = method.getName();
+                if (Modifier.isStatic(modifiers)) {
+                    LOG.info("invalid static @PostConstruct method {}.{}()", new Object[] { method.getDeclaringClass().getName(), methodName });
```
elandau: Why the newObject[]?

governator-core/src/main/java/com/netflix/governator/internal/PostConstructLifecycleActions.java
```diff
@@ -1,57 +1,119 @@
...
     }
-    
+
     @Override
     public String toString() {
         return "PostConstruct";
     }
 
+    private static class PostConstructVisitor implements TypeVisitor, Supplier<List<LifecycleAction>> {
+        private Set<String> visitContext = new HashSet<>();
```
elandau: 'seenMethods' is probably a better name

governator-core/src/main/java/com/netflix/governator/internal/PostConstructLifecycleActions.java
```diff
@@ -1,57 +1,119 @@
...
+        private LinkedList<LifecycleAction> typeActions = new LinkedList<>();
+
+        @Override
+        public boolean visit(final Class<?> clazz) {
+            return !clazz.isInterface();
```
elandau: Do we not want to support 'extends'?

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyLifecycleActions.java
```diff
@@ -1,56 +1,146 @@
...
         return "PreDestroy";
     }
 
+    private static class PreDestroyVisitor implements TypeVisitor, Supplier<List<LifecycleAction>> {
+        private Set<String> visitContext = new HashSet<>();
+        private List<LifecycleAction> typeActions = new ArrayList<>();
+
+        @Override
+        public boolean visit(final Class<?> clazz) {
+            boolean continueVisit = !clazz.isInterface();
```
elandau: nit: not of fan of tracking this boolean.  I'd rather see something like 

if (!clazz.isInterface())
   return false;

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyLifecycleActions.java
```diff
@@ -1,56 +1,146 @@
...
+        public String toString() {
+            return description;
+        }
+    }
+
+    private static final class AutoCloseableLifecycleAction implements LifecycleAction {
+        private final String description;
+
+        private AutoCloseableLifecycleAction(Class<? extends AutoCloseable> clazz) {
+            this.description = new StringBuilder().append("PreDestroy[").append(clazz.getName()).append("#")
```
elandau: s/PreDestroy/AutoCloseable?

governator-core/src/main/java/com/netflix/governator/internal/TypeInspector.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+    }
+
+    private static boolean _accept(Class<?> type, TypeVisitor visitor) {
+        if (type == null) {
+            return false;
+        }        
+        
+        boolean continueVisit = visitor.visit(type);
```
elandau: nit:  as before, I find the pattern of using a boolean here difficult to follow.  I'd much rather return when we know processing should stop.

elandau:
Has a few nits but overall looks good.  I like this approach much better.

governator-core/src/main/java/com/netflix/governator/internal/PostConstructLifecycleActions.java
topcat-70: without that Object[] ,there's some slf4j compile issue that manifests only
in the CI check; complains about 'java.lang.String' not being a 'Marker'.
Search of interwebs found the new Object[]{...} solution, which seems to
fix it.  Local gradle build worked fine both before and after the fix.

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyLifecycleActions.java
topcat-70: not a proponent of multiple exit points in a method

governator-core/src/main/java/com/netflix/governator/internal/PostConstructLifecycleActions.java
topcat-70: 'extends' is supported but 'implements' is not; it's pretty strongly worded
in jsr-250:

*The interfaces implemented by a class never contribute annotations to the
class itself or any of its members.*

we should likely go a step further and enforce that a class only specifies
a single @PostConstruct, but the implementation is a bit more involved.
I'm deferring that enforcement for now.

governator-core/src/main/java/com/netflix/governator/internal/TypeInspector.java
elandau: To reiterate my concern with using a boolean, it results in cyclomatic complexity which hurts readability and testability.  I think something like this would be better.

```java

if (type == null) {
    return false;
}        
        
if (!visitor.visit(type)) {
    return false;
}

for (final Field field : type.getDeclaredFields()) {
    if (!visitor.visit(field)) {
        return false;
    }
}

for (final Method method : type.getDeclaredMethods()) {
    if (!visitor.visit(method)) {
       return false;
    }
}

return _accept(type.getSuperclass(), visitor);
```

PR #298, opened 2016-05-29
Remove a bunch of classes that are no longer in use at Netflix.

elandau:


governator-legacy/src/main/java/com/netflix/governator/guice/BootstrapModule.java
```diff
@@ -18,7 +18,9 @@
 
 /**
  * Abstraction for binding during the bootstrap phase
+ * @deprecated Use standard Guice modules instead
  */
+@Deprecated
```
twicksell: Should probably at least link to a wiki article or something on how we expect people to handle early initialization of "platformy" dependencies.

PR #299, opened 2016-06-01
ordering for lifecycle listeners, fix equality in SafeLifecycleListener

tcellucci:
LifecycleManager uses a HashSet for discovered LifecycleListeners, providing unpredictable invocation order for callback methods.  Changed over to LinkedHashset so that LifecycleListener callbacks are called in same order as the listeners were added.

Added reverse-ordering for calling LifecycleListener.onStopped(), to ensure that listeners have best opportunity to 'tear down' any changes made in LifecycleListener.onStarted().

### DIFF
--- a/governator-core/src/main/java/com/netflix/governator/LifecycleInjectorCreator.java
+++ b/governator-core/src/main/java/com/netflix/governator/LifecycleInjectorCreator.java
@@ -116,7 +116,7 @@ public class LifecycleInjectorCreator implements InjectorCreator<LifecycleInject
             return lifecycleInjector;
         }
         catch (Exception e) {
-            LOG.error(String.format("Failed to create injector - %s@%d", e.getClass().getSimpleName(), System.identityHashCode(e)), e);
+            LOG.error("Failed to create injector - {}@{}", e.getClass().getSimpleName(), System.identityHashCode(e), e);
             onFailedInjectorCreate(e);
             try {
               manager.notifyStartFailed(e);
--- a/governator-core/src/main/java/com/netflix/governator/LifecycleInjectorCreator.java
+++ b/governator-core/src/main/java/com/netflix/governator/LifecycleInjectorCreator.java
@@ -116,10 +116,10 @@ public class LifecycleInjectorCreator implements InjectorCreator<LifecycleInject
             return lifecycleInjector;
         }
         catch (Exception e) {
-            LOG.error("Failed to create injector", e);
+            LOG.error(String.format("Failed to create injector - %s@%d", e.getClass().getSimpleName(), System.identityHashCode(e)), e);
             onFailedInjectorCreate(e);
             try {
-                manager.notifyStartFailed(e);
+              manager.notifyStartFailed(e);
             }
             catch (Exception e2) {
                 LOG.error("Failed to notify injector creation failure", e2 );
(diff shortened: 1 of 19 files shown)


governator-core/src/main/java/com/netflix/governator/LifecycleManager.java
```diff
@@ -60,17 +62,19 @@ public synchronized void notifyStarted() {
     public synchronized void notifyStartFailed(Throwable t) {
         if (state.compareAndSet(State.Starting, State.Stopped)) {
             failureReason = t;
-            for (LifecycleListener listener : listeners) {
-                listener.onStopped(t);
+            LifecycleListener[] asArray = listeners.toArray(new LifecycleListener[0]);
```
elandau: nit: I think the code would read better if we use an iterator instead of dealing with raw arrays. 

```java
Iterator<String> iter = new LinkedList<>(listeners).descendingIterator();
while (iter.hasNext()) {
    iter.next().onStopped();
}
```

elandau:
LGTM.  had just one small nit :)

governator-core/src/main/java/com/netflix/governator/LifecycleManager.java
topcat-70: that would iterate in the wrong order; the array is present to reverse the
direction.

topcat-70: didn't see the 'descendingIterator()' part; I'll add that in

governator-core/src/main/java/com/netflix/governator/LifecycleInjectorCreator.java
```diff
@@ -116,10 +116,10 @@ protected void configure() {
             return lifecycleInjector;
         }
         catch (Exception e) {
-            LOG.error("Failed to create injector", e);
+            LOG.error(String.format("Failed to create injector - %s@%d", e.getClass().getSimpleName(), System.identityHashCode(e)), e);
```
elandau: Let's use SLF4J syntax instead of String.format

governator-core/src/main/java/com/netflix/governator/internal/PostConstructLifecycleActions.java
```diff
@@ -108,15 +108,26 @@ public boolean visit(Field field) {
...
         @Override
         public void call(Object obj)
                 throws IllegalAccessException, IllegalArgumentException, InvocationTargetException {
-            LOG.debug("invoking lifecycle action {}", description);
-            method.invoke(obj);
+            LOG.debug("calling action {}", description);
+            try {
+                method.invoke(obj);
+            } catch (InvocationTargetException ite) {
+                Throwable cause = ite.getCause();
```
elandau: Can this be put in a common utility class?

tcellucci:
found some opportunities on Friday while making the change you'd suggested.  There was an issue with @PreDestroy not getting invoked for instances that had been @PostConstruct'd, if any of the LifecycleListeners threw an exception from OnStarted.  That's been fixed.

Also worked on exception handling, so that the thrown exception won't be wrapped in an InvocationTargetException if it can be avoided.  This allows the unit test 'LifecycleModuleTest' to work again, in cases where it expected specific failure types like AssertionError or RuntimeException.  That unit test was expanded to cover both RuntimeExceptions and Errors thrown from each of the lifecycle methods: PostCreate, OnStarted, OnStopped, and PreDestroy.

Lastly, I made some logging changes after looking at the unit test output.  LogManager now reports its starting, started, and stopped state changes; LifecycleListeners indicate when / which exception is responsible for an onStopped invocation; some work was done to prevent exceptions from being logged multiple times (i.e., by each LifecycleListener).   A good portion of that work is subjective improvement, please let me know what you think...

PR #300, opened 2016-06-06
Test runner

twicksell:


governator-test-junit/build.gradle
```diff
@@ -1,6 +1,11 @@
...
 
 dependencies {
-    compile     project(':governator')
-    compile     "junit:junit:4.11"
+    compile    project(':governator')
+    compile    project(':governator-api')
+    compile    project(':governator-core')
+    compile    'com.netflix.archaius:archaius2-guice:2.0.6'
+    compile    'com.netflix.archaius:archaius2-test:2.0.6'
+    compile    'junit:junit:4.12'
```
tcellucci: probably need to add junit version placeholder, to ensure that we don't conflict with junit version defined at top-level

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,210 @@
...
+import com.netflix.governator.LifecycleInjector;
+import com.netflix.governator.providers.SingletonProvider;
+
+public class AnnotationBasedTestInjectorManager {
+
+    private final LifecycleInjector injector;
+    private final List<Object> mocksToReset = new ArrayList<>();
+    private final List<Module> modulesForTestClass = new ArrayList<>();
+    private final List<Module> overrideModules = new ArrayList<>();
+    private final List<Key<?>> spyTargets = new ArrayList<>();
```
tcellucci: isn't the 'new ArrayList<>()' initialization style specific to Java 8?  I'd think you'd need to use 'new ArrayList<Key<?>>()'

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,210 @@
...
+    private final SettableConfig classLevelOverrides = new DefaultSettableConfig();
+    private final SettableConfig methodLevelOverrides = new DefaultSettableConfig();
+    private final TestPropertyOverrideAnnotationReader testPropertyOverrideAnnotationReader = new TestPropertyOverrideAnnotationReader();
+    private TestCompositeConfig testCompositeConfig;
```
tcellucci: final?

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,210 @@
...
+                        overrideModules.add(spyModule);
+                    }
+                    return null;
+                }
+            });
+        }
+    }
+    
+    public void prepareConfigForTestClass(Class<?> testClass, Method testMethod) {
+        for(Class<?> parentClass : ClassUtils.getAllSuperclasses(testClass)) {
```
tcellucci: probably should iterate through the superclasses in reverse order, so that properties from the most senior parent are overlaid by those of the more junior parent

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,210 @@
...
+    private final List<Object> mocksToReset = new ArrayList<>();
+    private final List<Module> modulesForTestClass = new ArrayList<>();
+    private final List<Module> overrideModules = new ArrayList<>();
+    private final List<Key<?>> spyTargets = new ArrayList<>();
+    private final SettableConfig classLevelOverrides = new DefaultSettableConfig();
+    private final SettableConfig methodLevelOverrides = new DefaultSettableConfig();
+    private final TestPropertyOverrideAnnotationReader testPropertyOverrideAnnotationReader = new TestPropertyOverrideAnnotationReader();
+    private TestCompositeConfig testCompositeConfig;
+
+    public AnnotationBasedTestInjectorManager(Class<?> classUnderTest) {
```
tcellucci: the method names { getModulesForTestClass, getMocksForTestClass getSpiesForTargetKeys } all have void return type; usually 'get' methods return a type.  maybe prefix the method names with 'construct' or 'inspect' instead, so method name matches function?

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    @SuppressWarnings("rawtypes")
+    private void getMocksForAnnotatedFields(Class<?> testClass) {
+
+        for (Field field : testClass.getDeclaredFields()) {
+            if (field.isAnnotationPresent(ReplaceWithMock.class)) {
+                overrideModules.add(new MockitoOverrideModule(field.getAnnotation(ReplaceWithMock.class), field.getType()));
+            }
+            if (field.isAnnotationPresent(WrapWithSpy.class)) {
```
tcellucci: is this key unique enough if there are multiple fields with the same type, either in the same class or in class parents?

governator-test-junit/src/main/java/com/netflix/governator/guice/test/ModulesForTesting.java
```diff
@@ -0,0 +1,47 @@
...
+ *     }
+ * }
+ * </pre>
+ */
+@Documented
+@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+public @interface ModulesForTesting {
+    
+    Class<? extends Module>[] value() default {};
```
tcellucci: default methods == Java 8; is that ok?

twicksell: This is an annotation, default values in annotations are fine prior to jdk8.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: Done

governator-test-junit/build.gradle
twicksell: There isn't a junit version set globally, and junit on the inherited projects (core, api, legacy) is only in testCompile so it isn't coming into this project transitively. 

In the case of this project, its a test utility and its junit version will be brought to consumers transitively, so I went with 4.12 as it is the latest release for junit.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,210 @@
...
+        for (Object mock : mocksToReset) {
+            Mockito.reset(mock);
+        }
+    }
+
+    public void cleanupInjector() {
+        injector.shutdown();
+    }
+
+    private LifecycleInjector createInjector(List<Module> modules, List<Module> overrides) {
```
elandau: Is this method necessary?  This code is only called from one place.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
elandau: I'd rather not have a direct dependency on commons-lang for this one use of getAllSuperclasses.  We can just have our own version of it.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: This isn't initialized in a constructor sadly. Its part of the test lifecycle, so I can't mark it final.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: Agreed

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: Made protected. It was intended to allow for extension if for any reason we change or need to modify the injector creation.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: It was not, just added a change for that. Right now I'm supporting the Named qualifier via the @WrapWithSpy(name="foo"). At some point we'll probably need to support any annotation qualifier, but I don't want to tackle that just yet.

tcellucci:
it might be useful to expand Spies annotation to match against arbitrary @Qualifier annotations for extensibility: 

@interface WrapWithSpy {
String name() default "";
Qualifier[] qualifiers() default {};
}
 E.g., for some custom Qualifier '@Toxic' that marked several classes, you could have an @WrapWithSpy(qualifiers= { @Toxic } )

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
pstout: Hmm, the only assignment to this field that I can find is at line 55 in the constructor.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,223 @@
...
+    
+    public void prepareConfigForTestClass(Class<?> testClass, Method testMethod) {
+        for(Class<?> parentClass : getAllSuperClassesInReverseOrder(testClass)) {
+            classLevelOverrides.setProperties(testPropertyOverrideAnnotationReader.getPropertiesForAnnotation(parentClass.getAnnotation(TestPropertyOverride.class)));
+        }
+        classLevelOverrides.setProperties(testPropertyOverrideAnnotationReader.getPropertiesForAnnotation(testClass.getAnnotation(TestPropertyOverride.class)));
+        methodLevelOverrides.setProperties(testPropertyOverrideAnnotationReader.getPropertiesForAnnotation(testMethod.getAnnotation(TestPropertyOverride.class)));                    
+    }
+    
+    public void cleanUpMethodLevelConfig() {
```
pstout: Consider grouping this with the other two cleanup methods, cleanupInjector and cleanupMocks.  Also, you capitalize "Up" in this method name, but not in the other two methods.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /**
+     * Injects dependencies into the provided test object.
+     */
+    public void prepareTestFixture(Object testFixture) {
+        injector.injectMembers(testFixture);
+    }
+
+    public void cleanupMocks() {
```
pstout: Do you have concrete use cases for wanting to clean up just the config, just the mocks, or just the injector?  I'm wondering whether you would end up with less boilerplate in consuming classes by just having a single cleanup method that takes care of doing all three.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,223 @@
...
+import com.netflix.archaius.test.TestPropertyOverrideAnnotationReader;
+import com.netflix.governator.InjectorBuilder;
+import com.netflix.governator.LifecycleInjector;
+import com.netflix.governator.providers.SingletonProvider;
+
+public class AnnotationBasedTestInjectorManager {
+
+    private final LifecycleInjector injector;
+    private final List<Object> mocksToReset = new ArrayList<>();
+    private final List<Module> modulesForTestClass = new ArrayList<>();
```
pstout: The modulesForTestClass and overrideModules fields are only used in the constructor and the appropriate "inspect" method.  If the inspect methods returned the collection that they were building up, then these fields could be "localized" to the constructor.  You might then want to rename the "inspect<Blah>" methods back to "get<Blah>".

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,223 @@
...
+        moduleClasses.addAll(getModulesForAnnotatedClass(testClass));
+        for (Class<?> parentClass : getAllSuperClassesInReverseOrder(testClass)) {
+            moduleClasses.addAll(getModulesForAnnotatedClass(parentClass));
+        }
+        for (Class<? extends Module> moduleClass : moduleClasses) {
+            try {
+                modulesForTestClass.add(moduleClass.newInstance());
+            } catch (InstantiationException | IllegalAccessException e) {
+                try {
+                    Constructor<?> constructor = moduleClass.getDeclaredConstructors()[0];
```
pstout: Can you explain the purpose of this try-block?  The order of constructors is not defined, so marking the first one as public in no way means that you are changing the no-arg constructor, should one exist.  If you want to try to handle the situation where the no-arg constructor is not public, then you should iterate over the declared constructors looking for one that has a zero value for getParameterCount().

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,223 @@
...
+                public <T> Void visit(Binding<T> binding) {
+                    final Binding<T> finalBinding = binding;
+                    if (spyTargets.contains(binding.getKey())) {
+                        AbstractModule spyModule = new AbstractModule() {
+                            protected void configure() {
+                                final String finalBindingName = "Spied "
+                                        + (finalBinding.getKey().getAnnotation() != null
+                                                ? finalBinding.getKey().getAnnotation().toString() : "")
+                                        + finalBinding.getKey().getTypeLiteral();
+                                final Key<T> newUniqueKey = Key.get(finalBinding.getKey().getTypeLiteral(),
```
pstout: Consider lifting the value of finalBinding.getKey().getTypeLiteral() into a local variable that you use here and on the previous line.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,223 @@
...
+                                        Names.named(finalBindingName));
+                                finalBinding.acceptTargetVisitor(new CopyBindingTargetVisitor<>(binder().bind(newUniqueKey)));
+                                bind(finalBinding.getKey()).toProvider(new SingletonProvider<T>() {
+                                    @Inject
+                                    Injector injector;
+
+                                    protected T create() {
+                                        T t = (T) injector.getInstance(newUniqueKey);
```
pstout: Does this cast produce a warning?  If so, consider adding the appropriate local @SurpressWarning annotation.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,223 @@
...
+        final Annotation annotation = testClass.getAnnotation(ModulesForTesting.class);
+        if (annotation != null) {
+            return Arrays.asList(((ModulesForTesting) annotation).value());
+        } else {
+            return Collections.emptyList();
+        }
+    }
+
+    private void inspectMocksForTestClass(Class<?> testClass) {
+        getMocksForAnnotatedFields(testClass);
```
pstout: I would move this after the for loop so that processing order for all classes is from base class (Object) down to the test class.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,223 @@
...
+                                                ? finalBinding.getKey().getAnnotation().toString() : "")
+                                        + finalBinding.getKey().getTypeLiteral();
+                                final Key<T> newUniqueKey = Key.get(finalBinding.getKey().getTypeLiteral(),
+                                        Names.named(finalBindingName));
+                                finalBinding.acceptTargetVisitor(new CopyBindingTargetVisitor<>(binder().bind(newUniqueKey)));
+                                bind(finalBinding.getKey()).toProvider(new SingletonProvider<T>() {
+                                    @Inject
+                                    Injector injector;
+
+                                    protected T create() {
```
pstout: @Override?

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,223 @@
...
+    public void cleanUpMethodLevelConfig() {
+        testCompositeConfig.resetForTest();
+    }
+    private List<Class<?>> getAllSuperClassesInReverseOrder(Class<?> clazz) {
+        List<Class<?>> allSuperclasses = ClassUtils.getAllSuperclasses(clazz);
+        Collections.reverse(allSuperclasses);
+        return allSuperclasses;
+    }
+
+    private class MockitoOverrideModule<T> extends AbstractModule {
```
pstout: static?

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    private void inspectSpiesForTargetKeys(List<Element> elements) {
+        for (Element element : elements) {
+            element.acceptVisitor(new DefaultElementVisitor<Void>() {
+                @Override
+                public <T> Void visit(Binding<T> binding) {
+                    final Binding<T> finalBinding = binding;
+                    if (spyTargets.contains(binding.getKey())) {
+                        AbstractModule spyModule = new AbstractModule() {
```
pstout: Consider lifting this anonymous module out and naming it, akin to MockitoOverrideModule.  This block is moderately convoluted.

governator-test-junit/build.gradle
tcellucci: I thought all the tests in governator-* moved to JUnit from TestNG, and that in the future they might be using your annotations for unit testing.  I think there needs to be a global JUnit version, and it should be the same here as it is in 'testCompile'  configuration elsewere in the project family.  Otherwise we could end up fail our own tests for JUnit version drift.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/CopyBindingTargetVisitor.java
```diff
@@ -0,0 +1,81 @@
...
+import com.google.inject.spi.ConvertedConstantBinding;
+import com.google.inject.spi.ExposedBinding;
+import com.google.inject.spi.InstanceBinding;
+import com.google.inject.spi.LinkedKeyBinding;
+import com.google.inject.spi.ProviderBinding;
+import com.google.inject.spi.ProviderInstanceBinding;
+import com.google.inject.spi.ProviderKeyBinding;
+import com.google.inject.spi.UntargettedBinding;
+
+public class CopyBindingTargetVisitor<T> implements BindingTargetVisitor<T, Void> {
```
pstout: Do you really want this to be part of the supported public API?  If not, consider making it package private.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/ReplaceWithMock.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.mockito.Answers;
+
+/**
+ * Creates an override binding for the annotated type. 
+ * 
+ * Example:
+ * <pre>
+ * {@literal @}RunWith(GovernatorJunit4ClassRunner.class)
+ * {@literal @}ModulesForTesting({ SomeTestModule.class })
```
pstout: Do you need the ModuleForTesting annotation even when you are using ReplaceWithMock?

twicksell: Yes. ModulesForTesting is how you declare which modules should be used when creating the injector. Everything about injector creation is hidden behind these annotations, so without ModulesForTesting you would be testing an empty injector.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/CopyBindingTargetVisitor.java
twicksell: I expect we may reuse it in future testing utilities, but easy enough to keep it package private. Will do

governator-test-junit/src/main/java/com/netflix/governator/guice/test/WrapWithSpy.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+@Target(FIELD)
+@Retention(RUNTIME)
+@Documented
+public @interface WrapWithSpy {
+
+    /**
+     * The name of the binding you wish to wrap with a Spy.
+     */
+    String name() default "";
```
pstout: Consider providing an example with a name.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: It does not, which means the cast is no longer necessary. I went through and cleaned up those warnings and just forgot to remove the cast. Removed

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: Added

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: Handling for non-public constructors. I'll iterate and search for the zero arg constructor as suggested.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: Not static as there is state being tracked within this module. Not a great thing, but I needed to keep hold of the list of things being mocked for cleanup purposes.

PR #302, opened 2016-06-10
add default paths from which to serve static content when using JettyModule`

twicksell:


tcellucci:
some kind of multibind for the static resources would be nice, so that I don't end up with static content served by some random lib that happens to be in my classpath, or such that I can have different content served on specific ports.  What do you think?

twicksell:
@tcellucci After thinking on it a change like that would belong elsewhere in governator. My intent here is to make sure META-INF/resources is served as per the servlet spec. This brings use of the JettyModule in line with the behavior a user would see running a war deployed to a servlet container. Reducing the change to only META-INF/resources so that the two behaviors are consistent. 

We can look at adding another default servlet elsewhere in governator to serve the other static resource paths, and make that configurable.

PR #305, opened 2016-06-23
Fix Jetty META-INF/resources discovery issue

drtechniko:
When multiple META-INF/resources directories existed in
a webapp we only picked the first one resolved.
Now we use the jetty-webapp tooling to scan for all resources.


drtechniko:
@twicksell Here we go

twicksell:
lgtm

PR #307, opened 2016-07-06
governator scope-aware predestroy handling

tcellucci:
introduce scope-aware lifecycle / predestroy handling in both governator-core and governator-legacy.

### DIFF
--- a/governator-core/src/main/java/com/netflix/governator/ManagedInstanceAction.java
+++ b/governator-core/src/main/java/com/netflix/governator/ManagedInstanceAction.java
@@ -1,0 +1,39 @@
+package com.netflix.governator;
+
+import java.lang.ref.Reference;
+import java.util.concurrent.Callable;
+
+/**
+ * Runnable that applies one or more LifecycleActions to a managed instance T or Reference<T>.
+ * For Reference<T> the action is invoked on a best-effort basis, if the referent is non-null
+ * at time run() is invoked
+ */
+public final class ManagedInstanceAction implements Callable<Void> {
+    private final Object target; // the managed instance
+    private final Reference<?> targetReference; // reference to the managed instance
+    private final Iterable<LifecycleAction> actions; // set of actions that will
+                                                     // be applied to target
+
+    public ManagedInstanceAction(Object target, Iterable<LifecycleAction> actions) {
+        this.target = target; // keep hard reference to target
+        this.targetReference = null;
+        this.actions = actions;
+    }
+
+    public ManagedInstanceAction(Reference<?> target, Object context, Iterable<LifecycleAction> actions) {
+        this.target = null;
+        this.targetReference = target; // keep hard reference to target
+        this.actions = actions;
+    }
+
+    @Override
+    public Void call() throws Exception {
+        Object target = (targetReference == null) ? this.target : targetReference.get();
+        if (target != null) {
+            for (LifecycleAction m : actions) {
+                m.call(target);
+            }
+        }
+        return null;
+    }
+}
--- a/governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
+++ b/governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
@@ -19,6 +19,10 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.inject.Binding;
 import com.google.inject.Key;
 import com.google.inject.Provider;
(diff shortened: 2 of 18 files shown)


governator-core/src/main/java/com/netflix/governator/ManagedInstanceAction.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.lang.ref.Reference;
+import java.util.concurrent.Callable;
+
+/**
+ * Runnable that applies one or more LifecycleActions to a managed instance T or Reference<T>.
+ * For Reference<T> the action is invoked on a best-effort basis, if the referent is non-null
+ * at time run() is invoked
+ */
+public final class ManagedInstanceAction implements Callable<Void> {
```
elandau: Any reason why this is a Callable<Void> and not a Runnable?

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -0,0 +1,220 @@
...
+     
+    private static class ScopeCleanupMarker {
+    }
+    
+    private final Deque<Callable<Void>> cleanupActions = new ConcurrentLinkedDeque<>();
+    
+    private final Map<Integer, ScopeCleanupAction> scopedCleanupActions = new LinkedHashMap<>();
+    private final Map<Class<? extends Annotation>, Scope> scopeBindings;
+    private final ReferenceQueue<ScopeCleanupMarker> markerReferenceQueue = new ReferenceQueue<>();
+    private final ExecutorService reqQueueExecutor = Executors.newSingleThreadExecutor();
```
elandau: Make this a deamon thread otherwise it'll block applications from shutting down.

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -0,0 +1,220 @@
...
+                            ScopeCleanupAction cleanupAction;
+                            synchronized(scopedCleanupActions) {
+                                cleanupAction = scopedCleanupActions.remove(markerKey);
+                            }
+                            if (cleanupAction != null) {
+                                cleanupAction.call();
+                            }
+                        }
+                    } 
+                    catch (InterruptedException e) {
```
elandau: Should we not terminate this thread if it's being interrupted?  And reset the interrupted state?

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -0,0 +1,220 @@
...
+    public <T> boolean register(T destroyableInstance, Object context, Iterable<LifecycleAction> action) {
+        return new ManagedInstanceScopingVisitor(destroyableInstance, context, action).visitEagerSingleton();
+    }
+    
+    /**
+     * final cleanup of managed instances if any
+     */
+    @Override
+    public synchronized void close() throws Exception {
+        if (!reqQueueExecutor.isShutdown()) reqQueueExecutor.shutdown();
```
elandau: To improve readability please put the call to reqQueueExecutor.shutdown(); on the next line.

governator-core/src/main/java/com/netflix/governator/ManagedInstanceAction.java
tcellucci: Runnable eats or wraps the exception in a RuntimeException; Callable.call is allowed to throw Exception 'as-is'.

governator-core/src/test/java/com/netflix/governator/PreDestroyTest.java
```diff
@@ -275,17 +318,148 @@ protected void configure() {
...
+        Foo managedFoo = null;
+        try (LifecycleInjector injector = builder.createInjector()) {
+            threadLocalScope.enter();
+            managedFoo = injector.getInstance(Foo.class);
+            Assert.assertNotNull(managedFoo);
+            Assert.assertFalse(managedFoo.isShutdown());
+            threadLocalScope.exit();
+            
+            System.gc();
+            Thread.sleep(GC_SLEEP_TIME);
```
elandau: Use Mockito.timeout() instead of sleep

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -0,0 +1,220 @@
...
+            action.call();
+        }     
+        cleanupActions.clear();
+    }
+    
+    /**
+     * visits bindingScope of managed instance to set up an appropriate strategy for cleanup
+     *  
+     */
+    private final class ManagedInstanceScopingVisitor implements BindingScopingVisitor<Boolean> {
```
elandau: It took me a while to figure out that ManagedInstanceScopingVisitor was adding to cleanupActions.  I believe a cleaner implementation would be to have ManagedInstanceScopingVisitor implements BindingScopingVisitor<LifecycleAction>  (instead of Boolean) and return the action object or null if there is none.  This would greatly improve testability of ManagedInstanceScopingVisitor.

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
tcellucci: made change to exit if interrupted, that allows Executor.shutdownNow() to function properly

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
tcellucci: ok, done

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -0,0 +1,237 @@
...
+         * 
+         */
+        @Override
+        public Boolean visitScope(Scope scope) {
+            Provider<ScopeCleanupMarker> scopedMarkerProvider = scope.scope(Key.get(ScopeCleanupMarker.class), markerProvider);
+            ManagedInstanceAction instanceAction = new ManagedInstanceAction(injectee, lifecycleActions);
+            ScopeCleanupMarker marker = scopedMarkerProvider.get();
+            Integer markerKey = System.identityHashCode(marker);
+            synchronized(scopedCleanupActions) {
+                if (scopedCleanupActions.containsKey(markerKey)) {
```
elandau: Since we're on Java8 maybe we can start using things like map#compute

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+    
+    /**
+     * final cleanup of managed instances if any
+     */
+    @Override
+    public synchronized void close() throws Exception {
+        if (running.compareAndSet(true, false)) {
+            reqQueueExecutor.shutdown();
+            reqQueueExecutor.awaitTermination(2, TimeUnit.SECONDS);
```
elandau: Why not use the boolean returned by awaitTermination?

elandau: 2 seconds seems very arbitrary and likely to fail for any codebase that has a lengthy shutdown.  We should either make this a configuration option, a much larger number (maybe 10 minutes) or just block indefinitely.

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -0,0 +1,237 @@
...
+            return true;
+        }
+
+        /*
+         * use ScopeCleanupMarker dereferencing strategy to detect scope closure, add new entry to scopedCleanupActions map
+         * 
+         */
+        @Override
+        public Boolean visitScope(Scope scope) {
+            Provider<ScopeCleanupMarker> scopedMarkerProvider = scope.scope(Key.get(ScopeCleanupMarker.class), markerProvider);
```
elandau: Key.get(ScopeCleanupMarker.class) can be a constant

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
tcellucci: waiting is optional, to allow the single reqQueue thread an opportunity to
finish processing a single PreDestroy method invocation.  Worst case
scenario for the 2-second wait is that a single PreDestroy method
invocation could be interrupted, before the code below this block invokes
the rest of them.  Long / indefinite waits would potentially delay /
completely block this process.

tcellucci: ok, sounds reasonable.  Will likely move to the end of the block as well

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
tcellucci: it's been a major pain to stick with Java 7 for this (all those inner
classes); seems like CI or some other build constraint is not accepting
Java 8.  You're saying Governator is all Java 8 currently?

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
tcellucci: will do

tcellucci:
resubmitting as single PR

PR #309, opened 2016-07-11
fix NPE for @Nullable injection

tcellucci:


governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -99,14 +99,24 @@ public PreDestroyMonitor(Map<Class<? extends Annotation>, Scope> scopeBindings)
     }
     
     public <T> boolean register(T destroyableInstance, Binding<T> binding, Iterable<LifecycleAction> action) {
-        return (running.get()) ? binding.acceptScopingVisitor(new ManagedInstanceScopingVisitor(destroyableInstance, binding.getSource(), action)) : false;
+        try {
+            return (running.get()) ? binding.acceptScopingVisitor(new ManagedInstanceScopingVisitor(destroyableInstance, binding.getSource(), action)) : false;
+        } catch (Exception e) {
+            LOGGER.warn("failed to register lifecycle actions for {}", destroyableInstance, e);
+            return false;
```
elandau: Shouldn't we be throwing here?   Under what conditions would this throw an exception that we are fine with ignoring?

PR #313, opened 2016-07-19
Deprecate all LifecycleInjectorBuilder methods that take module classes instead of instances

elandau:


twicksell:
+1

governator-legacy/src/main/java/com/netflix/governator/guice/LifecycleInjectorBuilder.java
```diff
@@ -96,6 +97,9 @@
      *
      * @param mainModule root application module
      * @return this
+     * 
+     * @deprecated 2016-07-19 This method of adding modules has been deemed an anti-pattern.  Use
+     * module instances with hashCode() and equals() for deduping.
```
qiangdavidliu: Given that the hashCode() and equals() implementations for deduping in modules are consistent, it would be useful to provide a code snippet as an example in these comments.

PR #314, opened 2016-07-21
upgrade to guice 4.1

twicksell:
address an edge case in which SingletonScope was not properly being provided with a reference to the Injector

### DIFF
--- a/governator-core/src/main/java/com/google/inject/internal/SingletonScopeInjectorProvider.java
+++ b/governator-core/src/main/java/com/google/inject/internal/SingletonScopeInjectorProvider.java
@@ -8,9 +8,9 @@ import com.google.inject.Injector;
  * A hack to provide a reference to the Injector to the {@link SingletonScope} for some
  * edge cases in which it was not properly being initialized. 
  */
-public class SingletonScopeInjectorProvider {
+public final class SingletonScopeInjectorProvider {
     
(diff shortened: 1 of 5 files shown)


governator-core/src/main/java/com/google/inject/internal/SingletonScopeInjectorProvider.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.lang.ref.WeakReference;
+
+import com.google.inject.Injector;
+
+/**
+ * A hack to provide a reference to the Injector to the {@link SingletonScope} for some
+ * edge cases in which it was not properly being initialized. 
+ */
+public class SingletonScopeInjectorProvider {
```
elandau: final?

governator-core/src/main/java/com/netflix/governator/LifecycleListenerModule.java
```diff
@@ -48,6 +52,7 @@ public String toString() {
 
         @Override
         public <T> void onProvision(ProvisionInvocation<T> provision) {
+            SingletonScopeInjectorProvider.setCurrentInjector(this.injector);
```
elandau: Rename to setIfAbsent

twicksell: done

PR #315, opened 2016-07-22
Fix NPE when calling .scope() on a singleton stop in PreDestroyMonitor

elandau:


governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -188,7 +194,13 @@ public Boolean visitEagerSingleton() {
          */
         @Override
         public Boolean visitScope(Scope scope) {
-            Provider<ScopeCleanupMarker> scopedMarkerProvider = scope.scope(MARKER_KEY, markerProvider);
+            final Provider<ScopeCleanupMarker> scopedMarkerProvider;
+            if (scope.equals(Scopes.SINGLETON) || scope.equals(LazySingletonScope.get()) || scope.equals(FineGrainedLazySingletonScope.get())) {
```
qiangdavidliu: It would be ideal if these scopes can be programmatically found instead of being hardcoded here.

PR #318, opened 2016-08-24
Builder provisioning

tcellucci:
adds composable pattern to Guice provisioning.  Intended for use with 'builder' style apis, to enable pluggable interaction with a builder before its end product is created / consumed.

Example:

```java
class SomeProviderModule extends AbstractModule {
        @ProvidesBuilder("myBuilder")
        public ListBuilder simpleProvider() {
            return new ListBuilder("one", "two", "three");
        }
    }
```
```java
 class ConsumerModule extends AbstractModule {
        @BuilderAdvice("myBuilder")       
        public Consumer<ListBuilder> fourFiveAndSixAdvice() {
            return list -> {
                list.add("four");
                list.add("five");
                list.add("six");
            };
        }
    }
```

```java
class InjectedBean {        
        private List<String> myList;
        @Inject
        public InjectedBean(ListBuilder listBuilder) {
            // expected list contains {one, two, three, four, five, six}
            this.myList = listBuilder.build();
        }
}
```


tcellucci:
note: PR is for Java 8, but governator currently builds with Java 7

tcellucci:
morphed into another PR

PR #319, opened 2016-08-31
Nested lifecycle listeners

tcellucci:
issue reported with nested lifecyclelisteners leads to iterator concurrent modification


elandau:
LGTM

PR #320, opened 2016-09-01
Add support for AutoBindSingleton

elandau:
Added a general ScanningModuleBuilder with support for class annotations on top of which AutoBindSingleton auto binding is implemented.  The resulting module can be installed directly on an injector.

```java
Guice.createInjector(new ScanningModuleBuilder().
     .forPackages("org.example")
     .addScanner(new AutoBindSingletonAnnotatedClassScanner())
     .build()
)
```

### DIFF
--- a/governator-core/build.gradle
+++ b/governator-core/build.gradle
@@ -7,6 +7,7 @@ dependencies {
     compile     "com.google.inject:guice:${guice_version}"
     compile     "com.google.inject.extensions:guice-multibindings:${guice_version}"
     compile     "com.google.inject.extensions:guice-grapher:${guice_version}"    // should be provided - only if you want graphing
+    compile     "org.ow2.asm:asm:5.0.4"
 
     testCompile "org.hamcrest:hamcrest-core:${hamcrest_version}"
     testCompile "org.hamcrest:hamcrest-library:${hamcrest_version}"
--- a/governator-core/build.gradle
+++ b/governator-core/build.gradle
@@ -7,7 +7,6 @@ dependencies {
     compile     "com.google.inject:guice:${guice_version}"
     compile     "com.google.inject.extensions:guice-multibindings:${guice_version}"
     compile     "com.google.inject.extensions:guice-grapher:${guice_version}"    // should be provided - only if you want graphing
-    compile     "org.ow2.asm:asm:5.0.4"
 
     testCompile "org.hamcrest:hamcrest-core:${hamcrest_version}"
     testCompile "org.hamcrest:hamcrest-library:${hamcrest_version}"
--- a/governator-core/build.gradle
+++ b/governator-core/build.gradle
@@ -7,7 +7,6 @@ dependencies {
     compile     "com.google.inject:guice:${guice_version}"
     compile     "com.google.inject.extensions:guice-multibindings:${guice_version}"
     compile     "com.google.inject.extensions:guice-grapher:${guice_version}"    // should be provided - only if you want graphing
-    compile     "org.ow2.asm:asm:5.0.4"
 
     testCompile "org.hamcrest:hamcrest-core:${hamcrest_version}"
     testCompile "org.hamcrest:hamcrest-library:${hamcrest_version}"
@@ -17,7 +16,4 @@ dependencies {
     test {
         useJUnit()
     }
-
-
-
 }
--- a/governator-core/build.gradle
+++ b/governator-core/build.gradle
@@ -7,6 +7,7 @@ dependencies {
     compile     "com.google.inject:guice:${guice_version}"
     compile     "com.google.inject.extensions:guice-multibindings:${guice_version}"
     compile     "com.google.inject.extensions:guice-grapher:${guice_version}"    // should be provided - only if you want graphing
+    compile     "org.ow2.asm:asm:5.0.4"
 
     testCompile "org.hamcrest:hamcrest-core:${hamcrest_version}"
     testCompile "org.hamcrest:hamcrest-library:${hamcrest_version}"
@@ -16,4 +17,7 @@ dependencies {
     test {
         useJUnit()
     }
+
+
+
 }
--- a/governator-core/src/main/java/com/netflix/governator/AutoBindSingletonAnnotatedClassScanner.java
+++ b/governator-core/src/main/java/com/netflix/governator/AutoBindSingletonAnnotatedClassScanner.java
@@ -21,6 +21,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
 import java.util.Set;
 
+import javax.inject.Scope;
 import javax.inject.Singleton;
 
 public class AutoBindSingletonAnnotatedClassScanner implements AnnotatedClassScanner {
@@ -59,8 +60,8 @@ public class AutoBindSingletonAnnotatedClassScanner implements AnnotatedClassSca
         
         Class<?> annotationBaseClass = getAnnotationBaseClass(annotation);
         
-        // AutoBindSingleton.class is used as a marker to mean "default" because annotation defaults cannot be null
-        if ( annotationBaseClass != AutoBindSingleton.class ) {
+        // Void.class is used as a marker to mean "default" because annotation defaults cannot be null
+        if ( annotationBaseClass != Void.class ) {
             Object foundBindingClass = searchForBaseClass(clazz, annotationBaseClass, Sets.newHashSet());
             Preconditions.checkArgument(foundBindingClass != null, String.format("AutoBindSingleton class %s does not implement or extend %s", clazz.getName(), annotationBaseClass.getName()));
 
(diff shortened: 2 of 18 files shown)


governator-core/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
     compile     "com.google.inject:guice:${guice_version}"
     compile     "com.google.inject.extensions:guice-multibindings:${guice_version}"
     compile     "com.google.inject.extensions:guice-grapher:${guice_version}"    // should be provided - only if you want graphing
+    compile     "org.ow2.asm:asm:5.0.4"
```
tcellucci: asm is fragile since so many things use it; can we use a shaded version, in guice or elsewhere in our dependencies instead?

governator-core/src/main/java/com/netflix/governator/AutoBindSingletonAnnotatedClassScanner.java
```diff
@@ -0,0 +1,179 @@
...
+        if (singletonAnnotation == null) {
+            LOG.info("***** {} should also be annotated with @Singleton to ensure singleton behavior", clazz.getName());
+        }
+        
+        Class<?> annotationBaseClass = getAnnotationBaseClass(annotation);
+        
+        // AutoBindSingleton.class is used as a marker to mean "default" because annotation defaults cannot be null
+        if ( annotationBaseClass != AutoBindSingleton.class ) {
+            Object foundBindingClass = searchForBaseClass(clazz, annotationBaseClass, Sets.newHashSet());
+            Preconditions.checkArgument(foundBindingClass != null, String.format("AutoBindSingleton class %s does not implement or extend %s", clazz.getName(), annotationBaseClass.getName()));
```
tcellucci: maybe 'declaring class'?

governator-core/src/main/java/com/netflix/governator/AutoBindSingletonAnnotatedClassScanner.java
```diff
@@ -0,0 +1,179 @@
...
+            builder.asEagerSingleton();
+        } else  {
+            builder.in(LazySingletonScope.get());
+        }
+    }
+    
+    private boolean hasScopeAnnotation(Class<?> clazz) {
+        Annotation scopeAnnotation = null;
+        for (Annotation annot : clazz.getAnnotations()) {
+            if (annot.annotationType().isAnnotationPresent(ScopeAnnotation.class)) {
```
tcellucci: add support for javax.inject.Scope annotation

governator-core/src/main/java/com/netflix/governator/ScanningModuleBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+ * {@code 
+ * install(new ScanningModuleBuilder().
+ *      .forPackages("org.example")
+ *      .scanForAnnotatedClasses(new AutoBindSingletonAnnotatedClassScanner())
+ *      .build()
+ * }
+ * </pre>
+ */
+public class ScanningModuleBuilder {
+    private List<String> packages = new ArrayList<>();
```
tcellucci: Set<String> better here?

governator-core/src/main/java/com/netflix/governator/ScanningModuleBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+        return this;
+    }
+    
+    /**
+     * Specify an {@link AnnotatedClassScanner} to process classes from the set of specified packages.  
+     * 
+     * @param packages
+     * @return Builder for chaining
+     */
+    public ScanningModuleBuilder scanForAnnotatedClasses(AnnotatedClassScanner scanner) {
```
tcellucci: appears not to scan here, just adds the argument to the list of scanners.  use 'addScanner()'?

governator-core/src/main/java/com/netflix/governator/ScanningModuleBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+        scanners.add(scanner);
+        return this;
+    }
+    
+    /**
+     * Exclude bindings for any class matching the specified predicate.
+     * @param classes
+     * @return Builder for chaining
+     */
+    public ScanningModuleBuilder excludeClasses(Predicate<String> predicate) {
```
tcellucci: should be clearer what the predicate is receiving.  Is it the fully-qualified classname of the candidate to scan?

governator-core/src/main/java/com/netflix/governator/ScanningModuleBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+    public ScanningModuleBuilder excludeClasses(String... classes) {
+        return excludeClasses(Arrays.asList(classes));
+    }
+    
+    /**
+     * Exclude specific classes from the classpath scanning.
+     * @param classes
+     * @return Builder for chaining
+     */
+    public ScanningModuleBuilder excludeClasses(Collection<String> classes) {
```
tcellucci: should take a Set here imo

governator-core/src/main/java/com/netflix/governator/ScanningModuleBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+            @Override
+            public void configure() {
+                for ( String basePackage : packages )  {
+                    try {
+                        String basePackageWithSlashes = basePackage.replace(".", "/");
+                        for (URL url : Collections.list(classLoader.getResources(basePackageWithSlashes))) {
+                            try {
+                                if ( isJarURL(url)) {
+                                    String jarPath = url.getFile();
+                                    if ( jarPath.contains("!") ) {
```
tcellucci: need 'continue' if no '!'?

governator-core/src/main/java/com/netflix/governator/ScanningModuleBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+                                        jarPath = jarPath.substring(0, jarPath.indexOf("!"));
+                                        url = new URL(jarPath);
+                                    }
+                                    File file = ClasspathUrlDecoder.toFile(url);
+                                    try (JarFile jar = new JarFile(file)) {
```
tcellucci: better to consolidate all the urls pertaining to a single jar, open it once and read all the jar entries in one shot.

governator-core/src/main/java/com/netflix/governator/ScanningModuleBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+                    
+                    @Override
+                    public AnnotationVisitor visitAnnotation(String desc, boolean visible) {
+                        Type type = getType(desc);
+                        if (includeRule.apply(className)) {
+                            for (AnnotatedClassScanner scanner : scanners) {
+                                if (getType(scanner.annotationClass()).equals(type)) {
+                                    try {
+                                        Class<?> cls = Class.forName(className, false, classLoader);
+                                        scanner.applyTo(binder(), cls.getAnnotation(scanner.annotationClass()), Key.get(cls));
```
tcellucci: maybe defer class loading to the scanner; pass it the classname and the binder only

governator-core/src/test/java/com/netflix/governator/package1/AutoBindSingletonMultiBinding.java
```diff
@@ -0,0 +1,11 @@
+package com.netflix.governator.package1;
+
+import com.netflix.governator.annotations.AutoBindSingleton;
+
+import javax.inject.Singleton;
+
+@AutoBindSingleton(multiple=true, value=AutoBindSingletonInterface.class)
+@Singleton
+public class AutoBindSingletonMultiBinding implements AutoBindSingletonInterface {
```
tcellucci: seems redundant to force declaration of an autobindsingleton as a singleton.  What else would it be?

governator-core/build.gradle
twicksell: +1

governator-core/src/main/java/com/netflix/governator/ScanningModuleBuilder.java
elandau: I tried using the same convention as Guice's scanModulesForAnnotatedMethods but you're right the name doesn't make sense here.

governator-core/src/main/java/com/netflix/governator/ScanningModuleBuilder.java
elandau: nope.  If present the '!' needs to be stripped from the jar name so the jar can be opened.  So, either case needs to continue processing.

governator-core/src/test/java/com/netflix/governator/package1/AutoBindSingletonMultiBinding.java
elandau: @AutoBindSingleton isn't actually a scope.  So, using this class without auto binding will not have singleton behavior.  Unfortunately we can't make AutoBindSingleton a scope since there's already code out there that has both and having two scopes on a class is not allowed in Guice.

PR #321, opened 2016-09-06
Updated JettyModule to accept Guice-bound JettyConnectorProviders.

JackOfMostTrades:


governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyConnectorProvider.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.eclipse.jetty.server.Connector;
+import org.eclipse.jetty.server.Server;
+
+/**
+ * Implementations of this interface should return a connector which may be added to a Jetty server. The server is
+ * passed as an argument to the getConnector() argument so that implementations can utilize subclasses of Jetty's
+ * ServerConnector which requires the Server as a constructor argument.
+ */
+public interface JettyConnectorProvider {
```
twicksell: Why JettyConnectorProvider instead of just binding Connector itself?

JackOfMostTrades: The classdoc somewhat describes; the ServerConnector apparently needs server passed to the constructor, so I didn't see any way for external modules to construct the connectors without some sort of callback that provide a reference to the Server.

PR #324, opened 2016-09-28
Governator Jersey-Guice integration

elandau:
Fork the jersey-guice integration to gain more control over Jersey's ResourceConfig and NOT auto-register bound classes with @Path annotations to the Jersey context.

Usage,
```java
LifecycleInjector injector = InjectorBuilder.fromModules(
    new JerseyServletModule() {
        @Override
        protected void configureServlets() {
            install(new GovernatorJerseySupportModule());
            serve("/*").with(GovernatorServletContainer.class);
        }
        
        @Advises
        @Singleton
        @Named("governator")
        UnaryOperator<DefaultResourceConfig> getResourceConfig() {
            return config -> {
                Map<String, Object> props = config.getProperties();
                props.put(ResourceConfig.FEATURE_DISABLE_WADL, "false");
 
                config.getClasses().add(SampleResource.class);

                return config;
            };
        }
    };
}
```


governator-jersey/build.gradle
```diff
@@ -1,7 +1,11 @@
 apply plugin: 'java'
 
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
 dependencies {
     compile project(':governator-core')
+    compile project(':governator-providers')
```
twicksell: governator-jersey-providers? Not to be confused with Guice Providers

elandau: This is the new project for @ProvidesWithAdvice, has nothing to do with Jersey.

zarfide:
Like.

governator-jersey/src/main/java/com/netflix/governator/guice/jersey/GovernatorJerseySupportModule.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.inject.Singleton;
+import javax.servlet.ServletContext;
+
+/**
+ * Module enabling support for Jersey configuration via advisable bindings for {@link DefaultResourceConfig}.
+ * This module provides an alternative Guice integration based on jersey-guice {@link GuiceContainer} but with 
+ * less opinions, such as automatically adding any bound class with {@literal @}Path as a resources.
+ * 
+ * Applications are expected to customize {@link DefaultResourceConfig} to provide a whitelist of
+ * root resources to expose.  This can be done by adding the following {@link @Advises} method.
```
qiangdavidliu: More documentation on 'where' the below should be added?

governator-jersey/src/main/java/com/netflix/governator/guice/jersey/GovernatorJerseySupportModule.java
```diff
@@ -0,0 +1,80 @@
...
+ * 
+ * <pre>{@code
+@Advises
+@Singleton
+@Named("governator")
+UnaryOperator<DefaultResourceConfig> adviseWithMyApplicationResources() {
+    return defaultResourceConfig -> {
+        defaultResourceConfig.getClasses()
+            .addAll(Arrays.asList(MyApplicationResource.class));
+        return defaultResourceConfig;
```
qiangdavidliu: Users may still want to have the ability to scan their (specifically declared) packages. Looks like a way to do so would be something like
```
DefaultResourceConfig scanned = new PackagesResourceConfig("com.my.specific.package");
defaultResourceConfig.getClasses()
    .addAll(scanned.getClasses());
```
Perhaps provide documentation/easier api for this use case?

governator-jersey/src/main/java/com/netflix/governator/guice/jetty/GovernatorServletContainer.java
```diff
@@ -1,94 +1,23 @@
...
- *             .build());
- * }
- * }
- * </pre>
+ * @see GovernatorJerseySupportModule
  */
 @Singleton
-public final class GovernatorServletContainer extends GuiceContainer {
+@Deprecated
+public final class GovernatorServletContainer extends com.netflix.governator.guice.jersey.GovernatorServletContainer {
```
qiangdavidliu: Is there any dependency on this class today that cannot break? This is a breaking change.

governator-jersey/src/main/java/com/netflix/governator/guice/jersey/GovernatorJerseySupportModule.java
```diff
@@ -0,0 +1,80 @@
...
+ * 
+ * Applications are expected to customize {@link DefaultResourceConfig} to provide a whitelist of
+ * root resources to expose.  This can be done by adding the following {@link @Advises} method.
+ * 
+ * <pre>{@code
+@Advises
+@Singleton
+@Named("governator")
+UnaryOperator<DefaultResourceConfig> adviseWithMyApplicationResources() {
+    return defaultResourceConfig -> {
```
qiangdavidliu: Also add documentation on changing config properties, such as
```
Map<String, Object> props = defaultResourceConfig.getProperties();
props.put(ResourceConfig.FEATURE_DISABLE_WADL, "false");
```

elandau: This one might be confusing since it will be implicitly done by an internal @Advice that ties properties to archaius.  I'll make it clear that any change to DefaultResourceConfig will be picked up.

governator-jersey/src/main/java/com/netflix/governator/guice/jersey/GovernatorJerseySupportModule.java
elandau: I really don't think we should be documenting all of jersey here.

governator-jersey/src/main/java/com/netflix/governator/guice/jetty/GovernatorServletContainer.java
elandau: We do have an internal dependency for the new runtime-lifecycle but it can be easily fixed by binding to GovernatorServletContainer.  Making this fully backwards compatible would end up overly complicating the implementation, especially when the fix is easy and impact minimal.

PR #326, opened 2016-10-05
fix java8 and invokedynamic issues

tcellucci:


governator-legacy/src/main/java/com/netflix/governator/lifecycle/LifecycleMethods.java
```diff
@@ -311,22 +312,21 @@ public static void methodInvoke(Method method, Object target) throws InvocationT
...
-    @SuppressWarnings("unchecked")
     public static <T> T fieldGet(Field field, Object obj) throws InvocationTargetException, IllegalAccessException {
         MethodHandle[] fieldHandler = fieldHandlesMap.get(field);
         if (fieldHandler == null) {
             fieldHandler = updateFieldHandles(field);
         }
         if (fieldHandler != EMPTY_FIELD_HANDLES) {
             try {
-                return (T)fieldHandler[0].invoke(obj);
+                return Modifier.isStatic(field.getModifiers()) ? (T)fieldHandler[0].invoke() : (T)fieldHandler[0].bindTo(obj).invoke();
```
elandau: nit: let's be consistent with use of ternary operator vs if-else

PR #327, opened 2016-10-06
Lifecycle methods restore

tcellucci:
these methods restored:

    @Deprecated
    public Collection<Method> methodsFor(Class<? extends Annotation> annotation) {
        return Arrays.asList(annotatedMethods(annotation));
    }

    @Deprecated
    public Collection<Field> fieldsFor(Class<? extends Annotation> annotation) {
        return Arrays.asList(annotatedFields(annotation));
    }

    @Deprecated
    public <T extends Annotation> Collection<T> classAnnotationsFor(Class<T> annotation) {
        return Arrays.asList(classAnnotations(annotation));
    }


elandau:
lgtm

PR #328, opened 2016-10-10
fix static method invocation with invokedynamic

tcellucci:


governator-legacy/src/main/java/com/netflix/governator/lifecycle/LifecycleMethods.java
```diff
@@ -315,9 +315,15 @@ public static void methodInvoke(Method method, Object target) throws InvocationT
         MethodHandle handler = methodHandlesMap.get(method);
         if (handler != null) {
             try {
-                handler.invoke(target);
+                if (Modifier.isStatic(method.getModifiers())) {
+                    log.warn("static lifecycle method: " + method + ", target=" + target);
```
elandau: nit: Use {} for parameter replacements
```java
log.warn("Invalid of @PostConstruct on static method={}, target={}", method, target);
```

PR #330, opened 2016-10-12
Jetty archaius integration

twicksell:


governator-jetty/src/main/java/com/netflix/governator/guice/jetty/Archaius2JettyModule.java
```diff
@@ -0,0 +1,32 @@
+package com.netflix.governator.guice.jetty;
+
+import javax.inject.Singleton;
+
+import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
+import com.netflix.archaius.ConfigProxyFactory;
+
+public class Archaius2JettyModule extends AbstractModule {
```
elandau: make this class final

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyModule.java
```diff
@@ -19,6 +21,7 @@
 import org.eclipse.jetty.webapp.WebInfConfiguration;
 
 import org.eclipse.jetty.webapp.WebXmlConfiguration;
+import org.junit.Before;
```
elandau: Remove this line as we do not have a compile dependency on junit

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyModule.java
```diff
@@ -106,6 +113,31 @@ public JettyRunner(Server server, final LifecycleManager manager) {
...
+    }
+    
+    @Singleton
+    static class OptionalJettyConfig {
+        
+        @com.google.inject.Inject(optional=true)
+        private JettyConfig jettyConfig;
+        
+        public JettyConfig getJettyConfig() {
+            if(jettyConfig != null)
```
elandau: Would much rather use the ternary operator here.  Or even Optional, since we're on Java8 now.
```
Optional.ofNullable(jettyConfig).orElse(() -> new DefaultJettyConfig());
```

PR #331, opened 2016-10-13
Test runner enhancements 

twicksell:


governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -38,22 +39,28 @@
...
-    private TestCompositeConfig testCompositeConfig;
+    private final TestCompositeConfig testCompositeConfig;
 
     public AnnotationBasedTestInjectorManager(Class<?> classUnderTest) {
+        this.injectorCreationMode = getInjectorCreationModeForAnnotatedClass(classUnderTest);
         inspectModulesForTestClass(classUnderTest);
         inspectMocksForTestClass(classUnderTest);
         inspectSpiesForTargetKeys(Elements.getElements(modulesForTestClass));
         testCompositeConfig = new TestCompositeConfig(classLevelOverrides, methodLevelOverrides);
         overrideModules.add(new ArchaiusTestConfigOverrideModule(testCompositeConfig));
```
tcellucci: why do we add this every time?  If my test isn't using archaius?

governator-test-junit/README.md
```diff
@@ -0,0 +1,81 @@
...
+* Annotate dependencies with [@ReplaceWithMock](https://github.com/Netflix/governator/blob/master/governator-test-junit/src/main/java/com/netflix/governator/guice/test/ReplaceWithMock.java) to add an override binding for that dependency's type to a Mockito implementation. 
+* Annotate dependencies with [@WrapWithSpy](https://github.com/Netflix/governator/blob/master/governator-test-junit/src/main/java/com/netflix/governator/guice/test/WrapWithSpy.java) to wrap any binding with a Mockito spy.
+* Annotate parent classes, test classes, and/or test methods with [@TestPropertyOverride](https://github.com/Netflix/archaius/blob/2.x/archaius2-test/src/main/java/com/netflix/archaius/test/TestPropertyOverride.java) to set [Archaius2](https://github.com/Netflix/archaius/tree/2.x) property overrides for your Injector **(Note: You must include [ArchaiusModule](https://github.com/Netflix/archaius/blob/2.x/archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java) yourself in @ModulesForTesting to use this feature.)**
+
+# Dependency
+Add the following to your build.gradle
+```
+testCompile 'com.netflix.governator:governator-test-junit:latest.release
```
tcellucci: missing close quote

governator-test-junit/src/main/java/com/netflix/governator/guice/test/TestInjectorCreationMode.java
```diff
@@ -0,0 +1,6 @@
+package com.netflix.governator.guice.test;
+
+public enum TestInjectorCreationMode {
+    BEFORE_CLASS,
```
tcellucci: I'd probably match the JUnit style of enum values here: { BEFORE_CLASS, BEFORE } is a match for existing annotations @Before and @BeforeClass

governator-test-junit/src/main/java/com/netflix/governator/guice/test/TestInjectorCreationMode.java
```diff
@@ -0,0 +1,6 @@
+package com.netflix.governator.guice.test;
+
+public enum TestInjectorCreationMode {
```
tcellucci: dont think this enum needs the 'Test' prefix.  There's only one injector in this context, no need to narrow it down to the 'Test' one.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/ModulesForTesting.java
```diff
@@ -43,5 +43,7 @@
 public @interface ModulesForTesting {
```
tcellucci: looking at the style of other JUnit annotations (@Test, @Before, @BeforeClass, @After, @ReplaceWithMock, etc), it's conspicuous to me that only @Test uses the 'test' word at all.  To cleanly extend their style with new annotations, I'd go with maybe @InjectorBuilder(modules=""), and @PropertyOverride rather than @TestPropertyOverride

governator-test-junit/src/test/java/com/netflix/governator/guice/test/junit4/GovernatorJunit4ClassRunnerArchaius2IntegrationTest.java
```diff
@@ -1,15 +1,20 @@
...
+import javax.inject.Named;
+import javax.inject.Singleton;
 
 import org.junit.FixMethodOrder;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.MethodSorters;
 
 import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
```
tcellucci: use javax provider if using javax Named and Singleton

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: It should be a harmless addition. I can add checks to see if its necessary or not but I figured that wasn't worth the added complexity.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/TestInjectorCreationMode.java
twicksell: I matched the Spring equivalent enum. Plus I did like that BEFORE_EACH_TEST_METHOD was very explicit about when the Injector was being created. BEFORE is a little ambiguous. Also worth noting that this is intended to be completely separate of JUnit as we will want to support TestNG, Spock, and Junit5 as well. Only the JUnit4Runner is tied to JUnit classes and conventions.

governator-test-junit/src/test/java/com/netflix/governator/guice/test/junit4/GovernatorJunit4ClassRunnerArchaius2IntegrationTest.java
twicksell: its the @Provides annotation, not Provider. No javax equivalent

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -90,16 +105,31 @@ private void inspectModulesForTestClass(Class<?> testClass) {
                 modulesForTestClass.add(moduleClass.newInstance());
             } catch (InstantiationException | IllegalAccessException e) {
                 try {
-                    Constructor<?> constructor = moduleClass.getDeclaredConstructors()[0];
-                    constructor.setAccessible(true);
-                    modulesForTestClass.add((Module) constructor.newInstance());
+                    Constructor<?>[] declaredConstructors = moduleClass.getDeclaredConstructors();
```
elandau: You can skip this loop by calling

```
Constructor<?> ctor = moduleClass.getDeclaredConstructor()
```

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -90,16 +105,31 @@ private void inspectModulesForTestClass(Class<?> testClass) {
...
+                    for(int i=0; i < declaredConstructors.length; i++) {
+                        if(declaredConstructors[i].getParameterCount() == 0) {
+                            Constructor<?> zeroArgConstructor = declaredConstructors[i];
+                            zeroArgConstructor.setAccessible(true);
+                            modulesForTestClass.add((Module) zeroArgConstructor.newInstance());
+                            break;
+                        }
+                    }
```
elandau: If there's not a no-arg constructor this class will be silently skipped.  We should log an error or propagate the exception from getDeclaredConstructor() in my comment above.

governator-test-junit/README.md
```diff
@@ -0,0 +1,81 @@
...
+    @Test
+    @TestPropertyOverride(propertyFiles={"testProps.properties"}) //Properties may also be loaded from a file
+    public void testConfigFromFile() {
+        assertEquals("test", config.getString("myProperty"));   //Verify that your property was set as expected
+    }
+}
+```
+
+# Controlling Injector Creation
+You may choose whether the Injector will be created once per test class or once per test method by setting the injectorCreation attribute of [@ModulesForTesting](https://github.com/Netflix/governator/blob/master/governator-test-junit/src/main/java/com/netflix/governator/guice/test/ModulesForTesting.java). **By default, the injector is created once per test class.** This should only be modified if your Injector is in some way stateful and that state has been modified by one of your tests.
```
noel-yap: What's actually important is whether or not the injected object is mutable. If the injected object is mutable, set the `injectorCreation` attribute of `@ModulesForTesting` to have a new instance of the `Injector` per test method.

PR #336, opened 2016-10-15
Provision metrics

elandau:
- Include @PostConstruct duration in provision metrics by adding the ProvisionMetricsListener prior to LifecycleProvisionListener in LifecycleInjectorCreator.
- Fix memory leak by no longer using thread locals in SimpleProvisionMetrics
- Stop collecting metrics once the injector has been created.


governator-core/src/main/java/com/netflix/governator/SimpleProvisionMetrics.java
```diff
@@ -1,28 +1,39 @@
...
 import javax.inject.Singleton;
 
-import com.google.inject.Key;
-
 @Singleton
-public class SimpleProvisionMetrics implements ProvisionMetrics {
-    private final ThreadLocal<Data> context = new ThreadLocal<Data>();
-    private final CopyOnWriteArraySet<Data> threads = new CopyOnWriteArraySet<>();
+public final class SimpleProvisionMetrics implements ProvisionMetrics {
+    private final ConcurrentMap<Long, Data> threads = new ConcurrentHashMap<>();
```
tcellucci: should size this hash map so multi-threaded access is not a bottleneck

governator-core/src/main/java/com/netflix/governator/SimpleProvisionMetrics.java
```diff
@@ -1,28 +1,39 @@
...
     
     public static class Data {
```
tcellucci: class name is disjoint from the ProvisioningMetrics function it facilitates; akin to 'Stuff' or 'Thing'. Also I would not expect to find methods in a class named 'Data', only 'data'.

governator-core/src/main/java/com/netflix/governator/ProvisionDebugModule.java
```diff
@@ -6,16 +6,16 @@
  * Install this module to log a Provision report after the Injector is created.
  */
 public final class ProvisionDebugModule extends SingletonModule {
```
tcellucci: seems more like a ProvisionReportModule if that's what it generates.  Seems like debugging is an unrelated or second-order concept

governator-core/src/main/java/com/netflix/governator/ProvisionMetricsModule.java
```diff
@@ -53,7 +66,20 @@ protected void configure() {
         this.bindListener(Matchers.any(), listener);
         requestStaticInjection(MetricsProvisionListener.class);
```
tcellucci: it feels hacky to use Guice's static injection feature solely to gain the side effect of early initialization.  Is there no other way?

elandau: nope.  unfortunately there is no other way, that I know of.  This pattern is used successfully by many other components.  Note that no statics are actually created, this just forces early initialization.

PR #338, opened 2016-10-18
ElementVisitor to output a warning about toInstance bindings

elandau:
toInstance() bindings risk forcing static initialization of code from the configure() method of a Guice module, which can have undersireable side effects for static bridges.  Adding ``.infoEachElement(new WarnOfToInstanceInjectionVisitor())`` to the InjectorBuilder can be used to identify toInstance() bindings.

```java
InjectorBuilder.fromModules(...)
   .infoEachElement(new WarnOfToInstanceInjectionVisitor())
   .createInjector();
```

### DIFF
--- a/governator-core/src/main/java/com/netflix/governator/visitors/WarnOfToInstanceInjectionVisitor.java
+++ b/governator-core/src/main/java/com/netflix/governator/visitors/WarnOfToInstanceInjectionVisitor.java
@@ -1,0 +1,25 @@
+package com.netflix.governator.visitors;
+
+import com.google.inject.Binding;
+import com.google.inject.spi.DefaultBindingTargetVisitor;
+import com.google.inject.spi.DefaultElementVisitor;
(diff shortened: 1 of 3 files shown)


governator-core/src/main/java/com/netflix/governator/visitors/WarnOfToInstanceInjectionVisitor.java
```diff
@@ -0,0 +1,18 @@
+package com.netflix.governator.visitors;
+
+import com.google.inject.Binding;
+import com.google.inject.spi.DefaultBindingTargetVisitor;
+import com.google.inject.spi.DefaultElementVisitor;
+import com.google.inject.spi.InstanceBinding;
+
+public class WarnOfToInstanceInjectionVisitor extends DefaultElementVisitor<String> {
```
twicksell: Class doesn't actually warn, so name seems a bit off.

PR #342, opened 2016-10-24
Enable provision metrics by default

elandau:
Metrics can be disable by adding the following binding

```java
bind(ProvisionMetrics.class).to(NullProvisionMetrics.class);
```

### DIFF
--- a/governator-core/src/test/java/com/netflix/governator/ProvisionMetricsModuleTest.java
+++ b/governator-core/src/test/java/com/netflix/governator/ProvisionMetricsModuleTest.java
@@ -16,7 +16,25 @@ import javax.inject.Singleton;
 
 public class ProvisionMetricsModuleTest {
     @Test
-    public void defaultMetricsAreEmpty() {
+    public void disableMetrics() {
+        try (LifecycleInjector injector = InjectorBuilder.fromModule(
+                new AbstractModule() {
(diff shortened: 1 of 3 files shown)


governator-core/src/test/java/com/netflix/governator/ProvisionMetricsModuleTest.java
```diff
@@ -30,13 +48,12 @@ protected void configure() {
...
                     }
                 })
-            .traceEachElement(new ProvisionListenerTracingVisitor())
             .createInjector()) {
             
             ProvisionMetrics metrics = injector.getInstance(ProvisionMetrics.class);
             LoggingProvisionMetricsVisitor visitor = new LoggingProvisionMetricsVisitor();
             metrics.accept(visitor);
-            Assert.assertTrue(visitor.getElementCount() == 0);
+            Assert.assertTrue(visitor.getElementCount() != 0);
```
tcellucci: what's this about?

elandau: The default behavior changed.  Previously it was a no-op.  Now the default is to collect metrics.  So the element count would be non-zero because Guice provisioned something and we tracked the metrics for it.

PR #343, opened 2016-10-24
Test runner

twicksell:


qiangdavidliu:
LGTM

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -48,20 +51,19 @@
     private final SettableConfig classLevelOverrides = new DefaultSettableConfig();
```
elandau: Do we really need to track these separately?  Why not just first populate class level overrides on a single SettableConfig and then apply the method overrides.

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -234,17 +234,48 @@ protected void configure() {
             bindingBuilder.toInstance(mock);
         }
     }
-    
+
     private static class ArchaiusTestConfigOverrideModule extends AbstractModule {
```
elandau: Instead of this entire module why not just have a new binding for @RuntimeLayer SettableConfig and override the core set of modules with this override?  Code that's injecting @RuntimeLayer SettableConfig and overriding configuration would just update the RuntimeLayer SettableConfig from the override module

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -234,17 +234,48 @@ protected void configure() {
...
+        @Provides
+        @Singleton
+        public TestCompositeConfig compositeConfig(OptionalConfigHolder config) {
+            return new TestCompositeConfig(config.get(), classLevelOverrides, methodLevelOverrides);
+        }
+
+        @Provides
+        @Singleton
+        @Raw
+        public CompositeConfig compositeConfig(TestCompositeConfig config) {
```
elandau: Why do we need this?  Can't we just continue to use the @Raw CompositeConfig from InternalArchaiusModule?

governator-test-junit/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: We reset class and method configs separately. We only reset the test config after individual tests.

PR #346, opened 2016-11-15
Create jersey root resources eagerly 

elandau:
Force create non Jersey Singleton annotated root resources eagerly when setting up Jersey so they are ready by the time the injector has been created.

### DIFF
--- a/governator-jersey/src/main/java/com/netflix/governator/guice/jersey/GovernatorServletContainer.java
+++ b/governator-jersey/src/main/java/com/netflix/governator/guice/jersey/GovernatorServletContainer.java
@@ -1,11 +1,5 @@
 package com.netflix.governator.guice.jersey;
 
-import com.google.inject.Injector;
-import com.sun.jersey.api.core.ResourceConfig;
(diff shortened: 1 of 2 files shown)


governator-jersey/src/main/java/com/netflix/governator/guice/jersey/GovernatorServletContainer.java
```diff
@@ -41,7 +45,20 @@ protected ResourceConfig getDefaultResourceConfig(
...
-        webapp.initiate(config, new GovernatorComponentProviderFactory(config, injector));
+        
+        GovernatorComponentProviderFactory factory = new GovernatorComponentProviderFactory(config, injector);
+        webapp.initiate(config, factory);
+        
+        // Make sure all root resources are created eagerly so they're fully initialized 
+        // by the time the injector was created, instead of being created lazily at the
+        // first request.
+        LOG.info("Creating resources : {}", config.getRootResourceClasses());
+        for (Class<?> resource : config.getRootResourceClasses()) {
```
noel-yap: How might this be tested?

PR #348, opened 2016-11-17
Change Grapher#graph roots check to be isEmpty instead of null

nathanschile:
Was getting a empty roots graph when using the Grapher class. `roots` can never be null, though it can be empty. Changed the graph method to check `isEmpty` on roots instead of null.


qiangdavidliu:
thanks @nathanschile

PR #351, opened 2016-11-24
Make ConfigurationLifecycleFeature lazy

elandau:
ConfigurationLifecycleFeature is initialized early as part of a static initialization path for framework components.  Since it injects ConfigurationProvider it can trigger early initialization of framework components (such as archaius) that are also dependent on static initialization for early instantiation.  This PR changes the behavior such that ConfigurationProvider is lazily instantiated when it is actually used.


tcellucci:
lotm

PR #355, opened 2017-02-01
fix out-of-order shutdown for eager singletons

tcellucci:


governator-core/src/test/java/com/netflix/governator/PreDestroyTest.java
```diff
@@ -172,6 +173,56 @@ public void shutdown2() {
...
+            protected void configure() {
+                bind(EagerBean.class).asEagerSingleton();
+                bind(SingletonBean.class).in(Scopes.SINGLETON);
+            }}).createInjector()) {
+            eagerBean = injector.getInstance(EagerBean.class);
+            singletonBean = injector.getInstance(SingletonBean.class);
+            Assert.assertFalse(eagerBean.shutdown);
+            Assert.assertFalse(singletonBean.shutdown);
+        }
+        Assert.assertTrue(eagerBean.shutdown);
```
elandau: Shouldn't the test confirm the expected shutdown ordering as well?

PR #358, opened 2017-03-10
New AdvisesBinder to programmatically create advises bindings

elandau:


governator-providers/src/test/java/com/netflix/governator/providers/AdvisesBinderTest.java
```diff
@@ -0,0 +1,158 @@
...
+    static class AdviseList implements UnaryOperator<List<String>> {
+        @Override
+        public List<String> apply(List<String> t) {
+            t.add("a");
+            return t;
+        }
+    }
+    
+    @Test
+    public void adviseWithoutAdvice() {
```
tcellucci: method name?  looks like this test does add bind advice

PR #360, opened 2017-03-28
Spock test framework

twicksell:


governator-test-spock/src/main/java/com/netflix/governator/test/GovernatorExtension.java
```diff
@@ -0,0 +1,68 @@
...
+import org.spockframework.runtime.model.SpecInfo;
+
+import com.netflix.governator.guice.test.AnnotationBasedTestInjectorManager;
+import com.netflix.governator.guice.test.InjectorCreationMode;
+import com.netflix.governator.guice.test.mocks.MockHandler;
+import com.netflix.governator.test.mock.spock.SpockMockHandler;
+
+import spock.lang.Specification;
+
+public class GovernatorExtension extends AbstractGlobalExtension {
```
elandau: I'm not familiar with Spock.  Can you add a description about what each of these implementation of Spock API's gives us.

governator-test/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -53,29 +54,30 @@
     private final TestPropertyOverrideAnnotationReader testPropertyOverrideAnnotationReader = new TestPropertyOverrideAnnotationReader();
     private TestCompositeConfig testCompositeConfig;
 
-    public AnnotationBasedTestInjectorManager(Class<?> classUnderTest) {
+    public AnnotationBasedTestInjectorManager(Class<?> classUnderTest, Class<? extends MockHandler> defaultMockHandlerClass) {
```
elandau: Is this a breaking change?  Are all tests now required to use spock?

governator-test/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
```diff
@@ -94,6 +96,20 @@ public InjectorCreationMode getInjectorCreationMode() {
     protected LifecycleInjector createInjector(List<Module> modules, List<Module> overrides) {
         return InjectorBuilder.fromModules(modules).overrideWith(overrides).createInjector();
     }
+    
+    protected MockHandler createMockHandlerForTestClass(Class<?> testClass, Class<? extends MockHandler> defaultMockHandler) {
```
elandau: Do we want to allow scanning of base classes as well or is using inheritance for tests an anti-pattern?

governator-test-spock/src/main/java/com/netflix/governator/test/GovernatorExtension.java
twicksell: Yeah, going to run through now and add javadoc and readmes. But this is a type for a ServiceLoader to find which binds our framework to the test class.

governator-test/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: Only breaking for people who've extended our test runner. Not breaking for people writing tests. And it does not require Spock. I added this type to let Spock users choose whether or not they wanted Spock mock handling or Mockito (as each offers a different set of pros and cons). It also serves as an extension point for introducing other Mock frameworks or customizing the behavior of existing ones.

governator-test/src/main/java/com/netflix/governator/guice/test/AnnotationBasedTestInjectorManager.java
twicksell: I should check the parent classes and override with any values given on child classes so that its consistent with all of the other annotation behavior. Good catch.

governator-test/src/main/java/com/netflix/governator/guice/test/mocks/MockHandler.java
```diff
@@ -0,0 +1,16 @@
+package com.netflix.governator.guice.test.mocks;
+
+
+/**
+ * Abstraction for creating/resetting Mocks and Spies. Must have a default constructor.
+ */
+public interface MockHandler {
```
tcellucci: Handler might not be the right noun here.  RequestHandler is something that processes api requests, but this MockHandler creates and resets mocks.  Should this be a Factory or Builder?

governator-test/src/main/java/com/netflix/governator/guice/test/mocks/MockHandler.java
```diff
@@ -0,0 +1,16 @@
...
+
+    <T> T createMock(Class<T> classToMock);
+
+    <T> T createMock(Class<T> classToMock, Object args);
```
tcellucci: is Object a good parameter type for 'args'?  usually I see it as Object[]

PR #364, opened 2017-05-01
remove old hibernate-validator dependency by removing rarely used

twicksell:
Validation support from governator-legacy

### DIFF
--- a/governator-legacy/src/test/java/com/netflix/governator/lifecycle/TestLifecycleManager.java
+++ b/governator-legacy/src/test/java/com/netflix/governator/lifecycle/TestLifecycleManager.java
@@ -18,7 +18,6 @@ package com.netflix.governator.lifecycle;
 
 import org.junit.Assert;
 import org.junit.Test;
-import javax.validation.constraints.Min;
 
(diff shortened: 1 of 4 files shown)


governator-legacy/src/test/java/com/netflix/governator/lifecycle/TestLifecycleManager.java
```diff
@@ -28,13 +27,13 @@ public void     testValidation() throws Exception
     {
         Object          goodObj = new Object()
         {
-            @Min(1)
+//            @Min(1)
```
elandau: why comment and not delete?

governator-legacy/src/test/java/com/netflix/governator/lifecycle/TestLifecycleManager.java
```diff
@@ -28,13 +27,13 @@ public void     testValidation() throws Exception
...
             private int     a = 10;
         };
 
         Object          badObj = new Object()
         {
-            @Min(1)
+//            @Min(1)
```
elandau: why comment and not release?

PR #365, opened 2017-05-15
always use Stage.TOOL when using ElementVisitors for inspection of bi…

twicksell:
…ndings


governator-servlet/src/main/java/com/netflix/governator/guice/servlet/WebApplicationInitializer.java
```diff
@@ -21,5 +23,9 @@
 public interface WebApplicationInitializer {
 
     Injector createInjector();
+    
+    default Injector createInjector(ServletContext context) {
```
elandau: Can you add some javadocs here.  Also, should createInjector be deprecated now?

elandau: Actually, where is this called?

PR #370, opened 2017-07-06
Ignore NoClassDefFoundException during governator classpath scanning

elandau:
Ignore NoClassDefFoundError during class path scanning. Real issues with missing jars will be reported when Guice actually provisions an object.

### DIFF
--- a/governator-legacy/src/main/java/com/netflix/governator/lifecycle/AnnotationFinder.java
+++ b/governator-legacy/src/main/java/com/netflix/governator/lifecycle/AnnotationFinder.java
@@ -113,7 +113,7 @@ public final class AnnotationFinder extends ClassVisitor {
                         annotatedFields.add(selfClass().getDeclaredField(name));
                         break;
                     } catch (NoSuchFieldException e) {
-                        throw new IllegalStateException("Error visiting field " + name, e);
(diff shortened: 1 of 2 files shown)


governator-legacy/src/main/java/com/netflix/governator/lifecycle/AnnotationFinder.java
```diff
@@ -97,9 +112,10 @@ public AnnotationVisitor visitAnnotation(String desc, boolean visible) {
                     try {
                         annotatedFields.add(selfClass().getDeclaredField(name));
                         break;
-                    } 
-                    catch (NoSuchFieldException e) {
-                        throw new IllegalStateException(e);
+                    } catch (NoSuchFieldException e) {
+                        throw new IllegalStateException("Error visiting field " + name, e);
+                    } catch (NoClassDefFoundError e) {
+                        log.info("Unable to scan field '{}' of class '{}'", name, selfClass().getName(), e.getMessage());
```
jeremycnf: Thanks, this will allow us to create lazy singleton providers with optional dependencies to reduce transitive dependency footprint while still fully participating in DI framework object management.

The addition of class name to the info message should be really helpful for people investigating unexpected behaviors. Would class name also be useful in the NoSuchFieldException message?

PR #392, opened 2019-04-20
fix memory leak for OptionalBinder with PreDestroy

brharrington:
If OptionalBinder is used with a type that has a PreDestroy
method, then PreDestroyMonitor will add a cleanup action each
time that type is injected. If this happens a lot, e.g. using
it with `RequestScoped`, then it can quickly fill up the heap
and cause OOM.

@tcellucci The `optional*` test cases reproduce the issue. Trying
your suggestion of just ignoring bindings with no scope seems to
have worked in that it passed the tests (I needed to add a scope to
one class for some Nullable tests). I don't know if it is the best way
to fix this issue, but it should avoid the memory leak.

/cc @vfilanovsky

### DIFF
--- a/governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
+++ b/governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
@@ -250,15 +250,14 @@ public class PreDestroyMonitor implements AutoCloseable {
             return rv;
         }
 
-        /*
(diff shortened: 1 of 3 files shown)


brharrington:
To keep the context in one place, @twicksell has some concerns about ignoring the cleanup for the unscoped objects. The Governator maintainers are investigating an alternative fix.

PR #393, opened 2019-04-25
Pre destroy no scope bug

tcellucci:
what is the problem: fix a long-standing memory leak caused by unscoped instances injected via Provider. The situation commonly occurs when a provider is declared for a type that implements AutoCloseable or is annotated with '@PreDestroy'.  At runtime, the injected instances are bound with 'no scope' by Guice, leading to a large number of cleanup actions pointing to potentially the same object.  The memory leak consists of the cleanup actions themselves, rather than the provided instance.

what is the fix: this PR adds specific handling of ProvidedInstanceBindings, to ensure that cleanup actions are not created every time a provider-sourced instance is injected.  It also refactors the PreDestroyMonitor class to hold any cleanup actions for 'unscoped' instances in a WeakHashMap, to ensure that only one cleanup action exists for an unscoped provisioned instance. 

The PR also introduces some code to log the number of unscoped instances present at shutdown, and shows where these instances were bound to the injector.

### DIFF
--- a/governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
+++ b/governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
@@ -7,6 +7,9 @@ import com.google.inject.Provider;
 import com.google.inject.Scope;
 import com.google.inject.Scopes;
 import com.google.inject.spi.BindingScopingVisitor;
+import com.google.inject.spi.DefaultBindingTargetVisitor;
+import com.google.inject.spi.Dependency;
+import com.google.inject.spi.ProviderInstanceBinding;
 import com.google.inject.util.Providers;
 import com.netflix.governator.LifecycleAction;
 import com.netflix.governator.ManagedInstanceAction;
@@ -17,7 +20,6 @@ import org.slf4j.LoggerFactory;
 import java.lang.annotation.Annotation;
 import java.lang.ref.Reference;
 import java.lang.ref.ReferenceQueue;
-import java.lang.ref.SoftReference;
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -25,6 +27,8 @@ import java.util.Deque;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.WeakHashMap;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
@@ -35,13 +39,15 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
- * Monitors managed instances and invokes cleanup actions when they become unreferenced
+ * Monitors managed instances and invokes cleanup actions when they become
+ * unreferenced
  * 
  * @author tcellucci
  *
  */
 public class PreDestroyMonitor implements AutoCloseable {
     private static Logger LOGGER = LoggerFactory.getLogger(PreDestroyMonitor.class);
+
     private static class ScopeCleanupMarker {
         static final Key<ScopeCleanupMarker> MARKER_KEY = Key.get(ScopeCleanupMarker.class);
         // simple id uses identity equality
@@ -62,7 +68,8 @@ public class PreDestroyMonitor implements AutoCloseable {
     }
 
     static final class ScopeCleaner implements Provider<ScopeCleanupMarker> {
-        ConcurrentMap<Object, ScopeCleanupAction> scopedCleanupActions = new ConcurrentHashMap<>(BinaryConstant.I14_16384);
+        ConcurrentMap<Object, ScopeCleanupAction> scopedCleanupActions = new ConcurrentHashMap<>(
+                BinaryConstant.I14_16384);
         ReferenceQueue<ScopeCleanupMarker> markerReferenceQueue = new ReferenceQueue<>();
         final ExecutorService reqQueueExecutor = Executors.newSingleThreadExecutor(
                 new ThreadFactoryBuilder().setDaemon(true).setNameFormat("predestroy-monitor-%d").build());
@@ -107,7 +114,8 @@ public class PreDestroyMonitor implements AutoCloseable {
         }
 
         /**
-         * Processes unreferenced markers from the referenceQueue, until the 'running' flag is false or interrupted
+         * Processes unreferenced markers from the referenceQueue, until the 'running'
+         * flag is false or interrupted
          * 
          */
         final class ScopedCleanupWorker implements Runnable {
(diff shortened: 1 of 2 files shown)


governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -133,33 +143,60 @@ public void run() {
...
     public PreDestroyMonitor(Map<Class<? extends Annotation>, Scope> scopeBindings) {
         this.scopeBindings = new HashMap<>(scopeBindings);
     }
 
     public <T> boolean register(T destroyableInstance, Binding<T> binding, Iterable<LifecycleAction> action) {
-        return scopeCleaner.isRunning() ? binding.acceptScopingVisitor(
-                new ManagedInstanceScopingVisitor(destroyableInstance, binding.getSource(), action)) : false;
+        if (scopeCleaner.isRunning()) {
+            boolean visitNoScope = Optional
+                    .ofNullable(binding.acceptTargetVisitor(new DefaultBindingTargetVisitor<T, Boolean>() {
```
elandau: How expensive is this operations?  Wondering if this should be moved into a cache.

elandau: To improve readability can you extract this anonymous class into a separate class.

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -133,33 +143,60 @@ public void run() {
...
     private ScopeCleaner scopeCleaner = new ScopeCleaner();
-
-    Map<Class<? extends Annotation>, Scope> scopeBindings;
+    private Map<Class<? extends Annotation>, Scope> scopeBindings;
 
     public PreDestroyMonitor(Map<Class<? extends Annotation>, Scope> scopeBindings) {
         this.scopeBindings = new HashMap<>(scopeBindings);
     }
 
     public <T> boolean register(T destroyableInstance, Binding<T> binding, Iterable<LifecycleAction> action) {
```
elandau: Looks like we check externally that `action` is not empty but might be good to either add a precondition check or return false if `action` is empty.

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -172,60 +209,80 @@ public void addScopeBindings(Map<Class<? extends Annotation>, Scope> bindings) {
...
     @Override
     public void close() throws Exception {
-        if (scopeCleaner.close()) { // executor thread to exit processing loop
+        if (scopeCleaner.close()) { // executor thread to exit processing loop            
             LOGGER.info("closing PreDestroyMonitor...");
-
-            for (Callable<Void> action : cleanupActions) {
-                action.call();
+            synchronized(cleanupActions) {
+                List<Map.Entry<Object, UnscopedCleanupAction>> actions = new ArrayList<>(cleanupActions.entrySet());
```
elandau: UnscopedCleanupAction already implements Comparable.  We can simplify this code using streams,
```
actions.entrySet()
		.stream()
		.sorted(Map.Entry.comparingByValue())
		.forEach(
				action -> action.getValue().call(action.getKey())
		);
```

governator-legacy/src/test/java/com/netflix/governator/guice/TestLifecycleInjector.java
```diff
@@ -11,98 +12,324 @@
...
+    Injector injector = LifecycleInjector.builder().withModules(m).build().createInjector();
 
     LifecycleManager mgr = injector.getInstance(LifecycleManager.class);
     mgr.start();
 
-    PreDestroyOnce obj = injector.getInstance(PreDestroyOnce.class);
-    for (int i = 0; i < 1000; ++i) {
-      PreDestroyOnce tmp = injector.getInstance(PreDestroyOnce.class);
-      Assert.assertSame(obj, tmp);
+    Object[] obj = new Object[25];
```
elandau: s/obj/objs

governator-legacy/src/test/java/com/netflix/governator/guice/TestLifecycleInjector.java
```diff
@@ -11,98 +12,324 @@
...
+    }
+
+    @PreDestroy
+    public void close() {
+      ++counter;
+      closed = true;
+    }
+
+    @Override
+    public int getCount() {
```
elandau: For improved readability please call this `getPreDestroyCallCount()`

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
tcellucci: can't do it with an Iterable without getting the iterator() first :(

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
tcellucci: hadn't seen Map.Entry.comparingByValue()

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
tcellucci: I extracted the anonymous class. Don't think the visitor is expensive; all the info is static structure / simple method calls

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
tcellucci: done

governator-legacy/src/test/java/com/netflix/governator/guice/TestLifecycleInjector.java
tcellucci: done

brharrington:
It seems like this is trying to treat objects with no scope as having singleton scope unless they happen to get collected prior to shutdown. If the objects are collected, then they will not get shutdown properly which could result in some sort of resource leak. So generally speaking the user cannot rely on the lifecycle management of objects with no scope. Perhaps it would be less confusing and more consistent overall if Governator never manages the lifecycle of these objects with no scope rather than partially doing it. Is there a known use-case where this partial handling is preferable?

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -242,29 +299,66 @@ public Boolean visitScopeAnnotation(final Class<? extends Annotation> scopeAnnot
...
+         * handle unscoped instances here using a 'best effort' strategy to clean up these instances iff they
+         * still exist when the PreDestroyMonitor is closed.
          */
         @Override
         public Boolean visitNoScoping() {
-            //cleanupActions.addFirst(
-            //        new ManagedInstanceAction(new SoftReference<Object>(injectee), context, lifecycleActions));
+            if (processNoScope) {
+                synchronized (cleanupActions) {
+                    cleanupActions.put(injectee, new UnscopedCleanupAction(context, lifecycleActions));
```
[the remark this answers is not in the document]
tcellucci: good find, I switched it over to a ConcurrentMap built with Guava's MapMaker().weakKeys(). 
 That means weak keys and identity-based equality checks, so issue should be resolved.

governator-core/src/main/java/com/netflix/governator/internal/PreDestroyMonitor.java
```diff
@@ -242,29 +299,66 @@ public Boolean visitScopeAnnotation(final Class<? extends Annotation> scopeAnnot
...
+    private static final class UnscopedCleanupAction implements LifecycleAction, Comparable<UnscopedCleanupAction> {
+        private volatile static long instanceCounter = 0;
+        private final long ordinal;
+        private final Object context;
+        private final Iterable<LifecycleAction> lifecycleActions;
+
+        public UnscopedCleanupAction(Object context, Iterable<LifecycleAction> lifecycleActions) {
+            this.context = context;
+            this.lifecycleActions = lifecycleActions;
+            this.ordinal = instanceCounter++;
```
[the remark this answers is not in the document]
tcellucci: I don't think that's an issue.  For guice instance provisioning, if the order of PreDestroy calls actually matters, then it's likely that the associated instances were created in the same thread (the jvm guarantees sequential same-thread ordering of operations on a volatile field, so the ordinal would be ever-increasing in that context).

governator-legacy/src/test/java/com/netflix/governator/guice/TestLifecycleInjector.java
```diff
@@ -11,98 +12,324 @@
...
       @Provides
-      public PreDestroyOnce providesObj() {
+      public PreDestroyOnce providesObj(String param1) {
         return new PreDestroyOnceImpl();
       }
-    });
+    }, PreDestroyOnce.class);
   }
 
+  /** registers ProviderInstanceBinding with 'no scope' 1000 times */
```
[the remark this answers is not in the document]
tcellucci: updated.

tcellucci:
performing cleanup on surviving 'no scope' instances during injector shutdown is a long-standing feature and needs to remain for backward compatibility.  
The reasoning at the time the feature was written is that Governator should make an effort to clean up any resources it creates, particularly since:
(1) the default Guice provisioning is 'unscoped' and 
(2) our home-built frameworks make heavy use of static internals that enable objects to outlive the injector
(3) we call 'PostConstruct' on unscoped instances. It was deemed 'least astonishing' to also call PreDestroy.

PR #395, opened 2019-06-07
Add finer-grained configuration options to governator-jetty.

JackOfMostTrades:
Adding more fiddly bits to Jetty configuration so I can drop other implementations that have copy/pasted and tweaked JettyModule.

### DIFF
--- a/governator-jetty/src/main/java/com/netflix/governator/guice/jetty/Archaius2JettyConfig.java
+++ b/governator-jetty/src/main/java/com/netflix/governator/guice/jetty/Archaius2JettyConfig.java
@@ -34,4 +34,12 @@ public interface Archaius2JettyConfig extends JettyConfig {
      */
     @DefaultValue("/")
     String getWebAppContextPath();
+
+    @DefaultValue("true")
+    boolean isPlaintextSocketEnabled();
+
+    @DefaultValue("16384")
+    int getRequestHeaderSize();
+
+    String getBindToHost();
 }
--- a/governator-jetty/src/main/java/com/netflix/governator/guice/jetty/Archaius2JettyConfig.java
+++ b/governator-jetty/src/main/java/com/netflix/governator/guice/jetty/Archaius2JettyConfig.java
@@ -36,10 +36,10 @@ public interface Archaius2JettyConfig extends JettyConfig {
     String getWebAppContextPath();
 
     @DefaultValue("true")
-    boolean isPlaintextSocketEnabled();
+    boolean isUnencryptedSocketEnabled();
 
     @DefaultValue("16384")
-    int getRequestHeaderSize();
+    int getRequestHeaderSizeBytes();
 
     String getBindToHost();
 }
--- a/governator-jetty/src/main/java/com/netflix/governator/guice/jetty/DefaultJettyConfig.java
+++ b/governator-jetty/src/main/java/com/netflix/governator/guice/jetty/DefaultJettyConfig.java
@@ -14,6 +14,12 @@ public class DefaultJettyConfig implements JettyConfig {
 
     private String webAppContextPath = "/";
 
+    private boolean plaintextSocketEnabled = true;
+
+    private int requestHeaderSize = 16384;
+
+    private String bindToHost = null;
+
     @Override
     public int getPort() {
         return port;
@@ -37,6 +43,21 @@ public class DefaultJettyConfig implements JettyConfig {
     @Override
     public String getWebAppContextPath() {
         return webAppContextPath;
+    }
+
+    @Override
+    public boolean isPlaintextSocketEnabled() {
+        return plaintextSocketEnabled;
+    }
+
+    @Override
+    public int getRequestHeaderSize() {
+        return requestHeaderSize;
+    }
+
+    @Override
+    public String getBindToHost() {
+        return bindToHost;
     }
 
     public DefaultJettyConfig setPort(int port) {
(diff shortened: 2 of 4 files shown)


governator-jetty/src/main/java/com/netflix/governator/guice/jetty/Archaius2JettyConfig.java
```diff
@@ -34,4 +34,12 @@
      */
     @DefaultValue("/")
     String getWebAppContextPath();
+
+    @DefaultValue("true")
+    boolean isPlaintextSocketEnabled();
```
tcellucci: plain text is a content type, but this boolean controls whether the socket uses encryption?  c/b isUnencryptedSocketEnabled()?

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyModule.java
```diff
@@ -82,6 +84,9 @@ private JettyConfig getConfig() {
  *
  */
 public final class JettyModule extends AbstractModule {
+
+    public static final String PLAINTEXT_CONNECTOR_NAME = "plaintext";
+
     private final static Logger LOG = LoggerFactory.getLogger(JettyModule.class);
```
tcellucci: no need for static here; Guice will ensure the singleton nature of this class and logger

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyModule.java
```diff
@@ -100,7 +105,14 @@ public JettyRunner(Server server, final LifecycleManager manager) {
...
                 server.start();
-                port = ((ServerConnector)server.getConnectors()[0]).getLocalPort();
+                int port = -1;
+                for (Connector connector : server.getConnectors()) {
+                    if (connector.getName().equals(PLAINTEXT_CONNECTOR_NAME)) {
+                        port = ((ServerConnector)connector).getLocalPort();
+                        break;
+                    }
+                }
+                this.port = port;
```
tcellucci: what happens if no connector found?  should we throw an exception?

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyModule.java
```diff
@@ -194,7 +206,7 @@ private Server getServer(OptionalJettyConfig optionalConfig, Set<JettyConnectorP
             webAppContext.setThrowUnavailableOnStartupException(true);
             webAppContext.addFilter(GuiceFilter.class, "/*", EnumSet.allOf(DispatcherType.class));
             webAppContext.addServlet(DefaultServlet.class, "/");
-            webAppContext.setResourceBase(config.getWebAppResourceBase());
+            webAppContext.setBaseResource(staticResourceBase);
```
tcellucci: not equivalent to previous behavior; with this change, we're restricting to classpath-only.  Is that purposeful?

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/DefaultJettyConfig.java
```diff
@@ -39,6 +45,21 @@ public String getWebAppContextPath() {
...
     }
 
+    @Override
+    public boolean isPlaintextSocketEnabled() {
+        return plaintextSocketEnabled;
+    }
+
+    @Override
+    public int getRequestHeaderSize() {
+        return requestHeaderSize;
```
tcellucci: c/b 'getRequestHeaderSizeBytes' if not a stylistic clash with other property names

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyModule.java
JackOfMostTrades: That was already there, but ok. :)

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/Archaius2JettyConfig.java
JackOfMostTrades: :+1:

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyModule.java
JackOfMostTrades: Yea, good call. Adding an exception.

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/DefaultJettyConfig.java
JackOfMostTrades: :+1:

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyModule.java
JackOfMostTrades: Actually, since the logger is used in the static JettyRunner inner class, it looks like this needs to be static. (And JettyRunner needs to be static since it's a Guice-bound class).

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyModule.java
JackOfMostTrades: This is the thing that I really wanted to understand a bit better. We get in this `if` block in the event that `staticResourceBase` is not null, but then that variable is never used anywhere. It seems like the intent would have been to use `staticResourceBase` here, but it's a bit unclear then what the `getWebAppResourceBase()` configuration was meant to control.

Personally, I just want to option to be able to use a classpath resource base instead of a file path resource base (which is what `setResourceBase` ends up giving). If you have some insight into how `staticResourceBase` and `webAppContextBase` are meant to interplay here maybe we could make that work in an intuitive way. But failing that, perhaps it should just be a new config flag whether we use `staticResourceBase` or `webAppContextBase`? Although in that case it's still not clear to be if this if block is checking the right thing.

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyModule.java
tcellucci: sorry didn't notice this was existing code until after I commented.  np to leave as-is.

governator-jetty/src/main/java/com/netflix/governator/guice/jetty/JettyModule.java
JackOfMostTrades: I just took a second pass at this. Rather than changing the behavior of staticResourceBase vs webAppResourceBase and disentangling what those mean, I made an update to the PR such that if webAppResourceBase starts with `classpath:` then it will treat it as a classpath resource. That should preserve backwards compatibility and be good enough for me.

PR #397, opened 2019-10-07
Add Javax dep for Java 11

carl-mastrangelo:


carl-mastrangelo:
Hi, this is needed to upgrade Netflix/zuul to Java 11.   The jar being added in was factored out of Java8 and into a separate package, thus the need to include the dep.

qiangdavidliu:
@carl-mastrangelo do you mind re pushing your two PRs? They are both failing on the builder JDK. The change to open should fix that. Thanks.

carl-mastrangelo:
@qiangdavidliu done PTAL.

Also as a side note, governator-test didn't compile for me.  The gradle file is missing several of the dependencies.

qiangdavidliu:
@asibross

carl-mastrangelo:
cc: @artgon

carl-mastrangelo:
Friendly ping, anything more to do here?

carl-mastrangelo:
FYI I updated a few more things to work with Java 11.   These should be pretty minor.  I confirmed this builds with both Java 11 and Java 8.

PTAL.   Also, I don't know what procedure is for merging, I assume you merge after you approve.

governator-jersey/build.gradle
```diff
@@ -14,6 +14,10 @@ dependencies {
 
     compileOnly 'javax.servlet:javax.servlet-api:3.0.1'
 
+    implementation "javax.xml.bind:jaxb-api:2.3.1"
```
asibross: I understand that jaxb was removed in Java 11, but where in governator code is it being used, or which of it's dependencies uses it?

governator-jersey/build.gradle
```diff
@@ -14,6 +14,10 @@ dependencies {
...
+    runtime "com.sun.xml.bind:jaxb-core:2.3.0.1"
+    runtime "com.sun.xml.bind:jaxb-impl:2.3.2"
```
asibross: If Jaxb is not being used but it's annotation are loaded when a class is loaded, it would suffice to just add the jaxb-api.
Also, an alternative to the split package jaxb implementation is `jaxb-runtime` from glassfish.
https://www.jesperdj.com/2018/09/30/jaxb-on-java-9-10-11-and-beyond/

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Thu Oct 10 13:23:52 PDT 2019
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-all.zip
```
asibross: Why is this change needed? *-all instead of *-bin.

governator-core/build.gradle
```diff
@@ -2,6 +2,7 @@ apply plugin: 'java'
 
 dependencies {
     compile     project(':governator-api')
+    compile     "javax.annotation:javax.annotation-api:1.3.2"
```
brharrington: This project is no longer maintained as it is part of the EE stuff that was transitioned to Eclipse:

https://github.com/javaee/javax.annotation/blob/master/README.md

The version of this package released by eclipse is:

https://mvnrepository.com/artifact/jakarta.annotation/jakarta.annotation-api/1.3.5

governator-jersey/build.gradle
carl-mastrangelo: Unit tests.  If you run the tests they pass, but complain about the missing classes.  I'm guessing the test is broken, but appears passing.

gradle/wrapper/gradle-wrapper.properties
carl-mastrangelo: This is what ./gradlew build did.  Not sure why.   Reverting it is probably a fruitless endeavor.

asibross:
Changes are covered by #400

PR #398, opened 2019-10-07
Bump Gradle and Nebula versions to latest

carl-mastrangelo:
In order to show the necessity of https://github.com/Netflix/governator/pull/397 , I tried building governator with Java11.   I was unable to do so because the version string parser couldn't parse JDK11 version:

```
* What went wrong:
Could not determine java version from '11.0.2'.
```


qiangdavidliu:
@carl-mastrangelo you probably want to do another push on this PR too to fix the oracle -> open JDK issues.

carl-mastrangelo:
@qiangdavidliu   Done!   BTW, You should probably test with multiple JDKs, which this change should allow us to do.

carl-mastrangelo:
cc: @artgon

qiangdavidliu:
@carl-mastrangelo looks like the build is still failing even with OpenJDK

brharrington:
Gradle 5 separates the compile and runtime scope of modules:

https://docs.gradle.org/5.0/userguide/java_library_plugin.html#sec:java_library_separation

You'll need to add a few direct dependencies to `governator-test` that were obtained transitively before. The dependency changes in https://github.com/Netflix/governator/pull/391 might still work.

carl-mastrangelo:
@brharrington @qiangdavidliu  Done, PTAL

PR #413, opened 2021-04-10
Avoid referring to internal Guice classes

carl-mastrangelo:
In Guice 4.2, guice-multibindings was merged into guice core, and in Guice 5, multibindings was deleted. When trying to upgrade to Guice5, I tried explicitly removing the multibindings dep as it is not rolled in, but some internal classes are no longer exposed.

In order to ease upgrading, Do a runtime check of the CircularDependency class, which was removed in Guice 5 ( https://github.com/google/guice/pull/1298/files ).


carl-mastrangelo:
FYI, only Guice 5 supports JDK's later than 11, so in order to update Our own JDK, we need to also update Guice, which in turn means governator must be updated.  Long term we would like to avoid Using governator/guice, but that is a much bigger change.  This is a surgical change to unblock me.

asibross (approved):

PR #419, opened 2023-03-17
Ignore AutoClosable interface on ExecutorService

DanielThomas:
JDK 19 adds AutoClosable to ExecutorService and existing injection points will not expect ExecutorService to be suddenly subject to PreDestroy lifecycle management.


DanielThomas:
@asibross just a quick heads up we'll need this for JDK 19 and later compatibility. I'll coordinate the release/integration if you're happy.

